PR #2, opened 2016-06-16
Tags and header

arpitux:
Fixed the overflowing tags and [+8] button,

Changed header structure to reduce one extra click to bring search box.

Added a Price filter UI. (In progress)


MicahZoltu:
Can you remove the pricing slider until it is ready?  When I merge this change it will go live immediately and I don't want users struggling to figure out how the slider works if it isn't fully implemented yet.

MicahZoltu:
Same with the reset button, can you hide it for now since it doesn't appear to work at the moment (or make it work) so I can get this merged?

MicahZoltu:
And can you also change the title to something like "Filter by Price" until the button updates with the real values.

arpitux:
I have commented out the complete Price filter as those predefined ranges are also not completely done.

PR #7, opened 2016-07-01
URL search

arpitux:
Search parameters get added in url.
Url with parameters can be used to see the same set of data.


js/functions.js
```diff
@@ -9,14 +9,14 @@ $(window).load(function () {
...
-        }
-    }
+    //if(!$('html').hasClass('ismobile')) {
+    //    if($('.page-loader')[0]) {
+    //        setTimeout (function () {
+    //            $('.page-loader').fadeOut();
+    //        }, 500);
+
+    //    }
+    //}
```
MicahZoltu, line 19: Recommend deleting this code instead of commenting it out if it isn't used anymore.

vendors/zoom-demo/imagezoom.js
```diff
@@ -118,6 +118,6 @@
   }
 
   $(document).ready(function(){
-    $('[data-imagezoom]').imageZoom();
+    //$('[data-imagezoom]').imageZoom();
```
MicahZoltu, line 121: Recommend deleting code instead of commenting it out.  Also, I think this whole function can go away rather than just its contents.

PR #13, opened 2016-07-30
Added Help tour and new loader

arpitux:
- Added a new button for a help which will start the help tour.
- New loader is added which will also appear when data is being changed (url hash change)
- Bugs fixed
  - Gray line coming over the search box


MicahZoltu:
👍

PR #15, opened 2016-08-11
Changes in Help Tour

arpitux:
- Removed additional steps from tour which are not required.
- Tour will start automatically for the first time user.
- During help tour, on adding a tag will now move the tour step forward.


MicahZoltu:
👍

PR #20, opened 2016-09-02
Implemented smart search

arpitux:
- added smart search where tags are chosen automatically when user enters different keywords and hit enters.
- Changes in Help tour


scripts/index.js
```diff
@@ -962,16 +961,19 @@ function ExecuteSmartSearch(txtbox) {
 
         SplitBackwardAndMatch(keyword_str);
 
+        console.log("search_matching_tags -");
+        console.dir(search_matching_tags);
```
MicahZoltu: Can you remove the logging before I merge this?  I don't want the users console to be filled up with non-error stuff.

PR #49, opened 2016-10-06
Side Fliter

arpitux:
- Added a side filter which shows Tags grouped by Category


client/scripts/index.js
```diff
@@ -452,6 +467,35 @@ Handlebars.registerHelper('titleCase', function (name) {
     return name.substr(0, 1).toUpperCase() + name.substr(1);
 });
 
+Handlebars.registerHelper('categoryIcon', function (name) {
+    var cssClass = "";
+    console.log(name);
```
MicahZoltu, line 472: This should be removed so as to not clutter the logs in production.

PR #58, opened 2016-10-26
MVC Razor

inabeel:
Added Mvc to support Dynamic meta tags to be rendered on server


inabeel:
Please verify

.gitignore
```diff
@@ -6,4 +6,3 @@
 /bin/
 /obj/
 
-/*.xproj.user
```
MicahZoltu, line 9: This shouldn't be removed.  `*.xproj.user` files are per-user settings, meaning yours may be different from mine.

Controllers/HomeController.cs
```diff
@@ -0,0 +1,43 @@
...
+        public IActionResult product(int productId)
+        {
+            ViewData["url"] = "Muhammad Nabeel";
+            ViewData["type"] = "Managing Director";
+            ViewData["title"] = "Principal Software Engineer / Technical Lead";
+            ViewData["description"] = "Relliks Systems";
+
+            return View("Index");
+        }
+    }
```
MicahZoltu: This project's code style is tab for indent, rather than spaces.  The easiest way to address this is to install the EditorConfig plugin for Visual Studio, which will follow the indentation style rules found in `.editorconfig` automatically.  Alternatively, you can change your Visual Studio settings to tabs, but you will have to switch it back and forth for every project you work on which can be a bit of a pain.

https://github.com/editorconfig/editorconfig-visualstudio#readme

Startup.cs
```diff
@@ -1,59 +1,91 @@
...
+			applicationBuilder.UseStaticFiles();
+
+            applicationBuilder.UseMvc(routes =>
+            {
+                routes.MapRoute(
+                    name: "default",
+                    template: "{controller=Home}/{action=Index}/{id?}");
+            });
+        }
+	}
```
MicahZoltu: This file appears to have mixed tabs and spaces, see comment above about EditorConfig plugin.

bags-client.sln
```diff
@@ -1,22 +1,22 @@
...
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{6BE8394A-B198-4B1B-80F5-8EB947F3C3F5}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{6BE8394A-B198-4B1B-80F5-8EB947F3C3F5}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{6BE8394A-B198-4B1B-80F5-8EB947F3C3F5}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{6BE8394A-B198-4B1B-80F5-8EB947F3C3F5}.Release|Any CPU.Build.0 = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(SolutionProperties) = preSolution
+		HideSolutionNode = FALSE
+	EndGlobalSection
+EndGlobal
```
MicahZoltu: It looks like your settings also are messing with the line-endings in files.  If you get the `.editorconfig` mentioned previously I believe it will do the right thing going forward.  In the meantime, it doesn't appear there are any actual changes in this file, so you can just remove it from the change set.

bags-client.xproj.user
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <ActiveDebugProfile>Kestrel</ActiveDebugProfile>
+  </PropertyGroup>
+</Project>
```
MicahZoltu: As mentioned previously, this should not be committed to source control.

client/scripts/templates/template.min.js
```diff
@@ -0,0 +1,254 @@
+(function() {
```
MicahZoltu: This file shouldn't be committed to source control.  I'll add it to the .gitignore so it doesn't accidentally get re-added in a future PR.

bags-client.sln
MicahZoltu: Actually, it appears this file may have had the wrong (or right depending on how you look at it) line endings previously... I'll investigate and figure out what the right answer is but in the meantime you should still remove this from the change set.

project.json
```diff
@@ -13,8 +14,14 @@
...
-		"Microsoft.Extensions.Logging.Console": "1.0.0"
-	},
+    "Microsoft.Extensions.Logging.Console": "1.0.0",
+    "Microsoft.AspNetCore.Mvc": "1.0.1",
+    "Microsoft.AspNetCore.Routing": "1.0.1",
+    "Microsoft.AspNetCore.Diagnostics": "1.0.0",
+    "Microsoft.VisualStudio.Web.BrowserLink.Loader": "14.0.0",
+    "Microsoft.AspNetCore.Razor.Tools": "1.0.0-preview2-final",
+    "Microsoft.Extensions.Configuration.FileExtensions": "1.0.0",
+    "Microsoft.Extensions.Configuration.Json": "1.0.0"	},
```
MicahZoltu: As previously, the indentation in this file is off.  This might have just been Visual Studio, it has a tendency to format this file wrong sometimes.  Either way, please switch it to tabs.

MicahZoltu: These two extensions shouldn't be necessary (`Json` and `FileExtensions`).  Similarly, I don't think `BrowserLink.Loader` is necessary nor is `Diagnostics`.  Recommend removing them and see if everything still works.

I'm not sure what `Razor.Tools` is for, I have a sneaking suspicion it isn't necessary either but I'm less confident.  Can you try removing it and seeing if things still work?

web.config
```diff
@@ -0,0 +1,14 @@
...
+    Configure your application settings in appsettings.json. Learn more at http://go.microsoft.com/fwlink/?LinkId=786380
+  -->
+
+  <system.webServer>
+    <handlers>
+      <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModule" resourceType="Unspecified"/>
+    </handlers>
+    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" forwardWindowsAuthToken="false"/>
+  </system.webServer>
+</configuration>
```
MicahZoltu: This file shouldn't be necessary since the app runs in Kestrel.  I believe this is only necessary for IIS hosted applications, so it should be safe to delete.

Startup.cs
```diff
@@ -1,59 +1,91 @@
...
+			applicationBuilder.UseApplicationInsightsExceptionTelemetry();
+			applicationBuilder.UseDefaultFiles();
+			applicationBuilder.UseStaticFiles();
+
+            applicationBuilder.UseMvc(routes =>
+            {
+                routes.MapRoute(
+                    name: "default",
+                    template: "{controller=Home}/{action=Index}/{id?}");
+            });
```
MicahZoltu: It appears that you are using attribute naming for routing, so I don't think this block is necessary anymore.  In fact, as it is setup right now it appears to add a route that isn't actually (shouldn't be) a valid route.  Can we remove this whole `UseMvc(...)` block and have everything still work?

Startup.cs
```diff
@@ -1,59 +1,91 @@
...
+
+            if (hostingEnvironment.IsDevelopment())
+            {
+                applicationBuilder.UseDeveloperExceptionPage();
+                applicationBuilder.UseBrowserLink();
+            }
+            else
+            {
+                applicationBuilder.UseExceptionHandler("/Home/Error");
+            }
```
MicahZoltu: All of this can be deleted I think.  At the least, the exception handler and developer exception page.  I have never been a fan of using browser link so I would recommend deleting that one as well for now.  We may bring the `BrowserLink` stuff back later, but at this phase I would prefer to keep things trim and simple and since UI development isn't being done inside the VS project at the moment it is unnecessary.

Startup.cs
```diff
@@ -1,59 +1,91 @@
...
+		{
+			loggerFactory.AddConsole(minLevel: LogLevel.Warning);
+
+            // custom middleware to re-route all paths starting with `/app` to `/index.html`
+            //applicationBuilder.Use(async (context, next) =>
+            //{
+            //	if (context.Request.Path.StartsWithSegments(new PathString("/app")))
+            //		context.Request.Path = new PathString("/index.html");
+            //	await next();
+            //});
```
MicahZoltu: Delete this code if it is unused.  However, I suspect this code (or something similar is actually necessary to get all of the routes hooked up.  At the moment it appears that the hooked up route is `[Route("app", Name = "app", Order = 0)]`, which I'm not sure will catch everything inside of `/app`.  We need the controller action to be called for every route that starts with `/app`, which includes `/app/aboutus`, `/app/foo`, etc.  We may break some of those out into separate manual routes later but for now we want a single catch-all route that handles everything starting with `/app`.

One way to achieve this may be to leave this middleware in place (uncomment it) and make `/index.html` be the only route handled by the controller action.  The middleware will trigger first, redirecting everything in `/app` to `/index.html` and then MVC will pick up that route and render it.

Startup.cs
```diff
@@ -1,59 +1,91 @@
...
+            services.AddMvc();
+
+            services.Configure<RazorViewEngineOptions>(options =>
+            {
+                var previous = options.CompilationCallback;
+                options.CompilationCallback = (context) =>
+                {
+                    previous?.Invoke(context);
+                };
+            });
```
MicahZoltu: I think this can be removed since we don't have any razor compilation settings to set.  At the moment, this code is just adding an unnecessary layer and confusing to the code.

I think you can delete the entire `services.Configure<RazorViewEngineOptions>(...)` block.

Startup.cs
```diff
@@ -1,59 +1,91 @@
...
+
+		private IConfiguration _configuration;
+
+		public Startup(IHostingEnvironment hostingEnvironment)
+		{
+			_configuration = new ConfigurationBuilder()
+				.AddApplicationInsightsSettings(developerMode: hostingEnvironment.IsDevelopment())
+                       .SetBasePath(hostingEnvironment.ContentRootPath)
+                .AddJsonFile("appsettings.json", optional: true, reloadOnChange: true)
+                .AddJsonFile($"appsettings.{hostingEnvironment.EnvironmentName}.json", optional: true)
```
MicahZoltu: This service only gets settings right now from environment variables, so we shouldn't be using JSON settings as it adds confusion to the code without benefit.  At the moment, I believe this service actually has no settings in fact.

Startup.cs
```diff
@@ -1,59 +1,91 @@
...
+				.Run();
+		}
+
+		private IConfiguration _configuration;
+
+		public Startup(IHostingEnvironment hostingEnvironment)
+		{
+			_configuration = new ConfigurationBuilder()
+				.AddApplicationInsightsSettings(developerMode: hostingEnvironment.IsDevelopment())
+                       .SetBasePath(hostingEnvironment.ContentRootPath)
```
MicahZoltu: `SetBasePath` can also be removed if we aren't doing JSON based service configuration.

Controllers/HomeController.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        public IActionResult Index()
+        {
+            ViewData["url"] = "https://bagcupid.com/";
+            ViewData["type"] = "website";
+            ViewData["title"] = "Bag Cupid";
+            ViewData["description"] = "What is your dream bag? Are you having trouble finding it? Let us help you!";
+
+            return View();
+        }
```
MicahZoltu: I think we can delete this.  The action that matters is the one below on line 22, this is just a duplicate.

Controllers/HomeController.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        [Route("app", Name = "app", Order = 0)]
+        public IActionResult app()
```
MicahZoltu: Style: Function names should start with a capital letter.  `App()` in this case.

Controllers/HomeController.cs
```diff
@@ -0,0 +1,43 @@
...
+        [Route("app/product/{productId}", Name = "product", Order = 0)]
+        public IActionResult product(int productId)
+        {
+            ViewData["url"] = "Muhammad Nabeel";
+            ViewData["type"] = "Managing Director";
+            ViewData["title"] = "Principal Software Engineer / Technical Lead";
+            ViewData["description"] = "Relliks Systems";
+
+            return View("Index");
+        }
```
MicahZoltu: This method appears to be test code, so it shouldn't be included in the code.  Please remove it from the change set.

Controllers/HomeController.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        [Route("app", Name = "app", Order = 0)]
+        public IActionResult app()
+        {
+            ViewData["url"] = "https://bagcupid.com/";
+            ViewData["type"] = "website";
+            ViewData["title"] = "Bag Cupid";
+            ViewData["description"] = "What is your dream bag? Are you having trouble finding it? Let us help you!";
+
+            return View("Index");
```
MicahZoltu: The way this is setup right now, it is going to look for `Views/Home/index.cshtml`.  I would like to figure out a way to make it so instead it gets its view from `client/index.html` instead, but parses it as a `cshtml` file.  This would make it so my client developers don't have to change their workflow in order to mess with this file.  This is the part that I don't know enough about Razor to do, and may require a little research to figure out.

Controllers/HomeController.cs
```diff
@@ -0,0 +1,43 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Zoltu.Bags.Client.Controllers
+{
+    public class HomeController : Controller
```
MicahZoltu: Lets rename this to `AppController`.

MicahZoltu (changes requested):
You are off to a good start.  A good portion of the feedback here is just code style stuff that will help keep the code in this repo consistent with other repos in this project.  This includes things like indentation (tabs instead of spaces), newlines (LF instead of CRLF) and what should/shouldn't be committed to the repository.  These changes should be easy, if you want some help with git magic to make them happen I don't mind helping out if you can catch me when I am at home (evenings and weekends PDT).

There is also a handful of recommendations on code that can be deleted.  I like to keep the code that is committed to the repository only the code strictly necessary to make things work.  As the project grows, it can quickly lead to confusion when there is code in the repository that is unused/unnecessary because it isn't always obvious which is which when troubleshooting an issue.

Finally, I did have some comments on the controller action and MVC setup itself.  See my comments in this review for details.

inabeel:
Dear Micah,

I have seen your comments and got your points. Most part of it is because of visual studio code settings

I will fix this stuff tonight and will let you know about it. We can also sit together on weekend to discuss all aspects of coding conventions you follow.

Thank you for your precious time.

Warm Regards,

Muhammad Nabeel

On Thu, Oct 27, 2016 at 6:42 AM +0500, "Micah Zoltu" <notifications@github.com<mailto:notifications@github.com>> wrote:

@MicahZoltu requested changes on this pull request.

You are off to a good start. A good portion of the feedback here is just code style stuff that will help keep the code in this repo consistent with other repos in this project. This includes things like indentation (tabs instead of spaces), newlines (LF instead of CRLF) and what should/shouldn't be committed to the repository. These changes should be easy, if you want some help with git magic to make them happen I don't mind helping out if you can catch me when I am at home (evenings and weekends PDT).

There is also a handful of recommendations on code that can be deleted. I like to keep the code that is committed to the repository only the code strictly necessary to make things work. As the project grows, it can quickly lead to confusion when there is code in the repository that is unused/unnecessary because it isn't always obvious which is which when troubleshooting an issue.

Finally, I did have some comments on the controller action and MVC setup itself. See my comments in this review for details.

---

In .gitignorehttps://github.com/Zoltu/bags-client/pull/58#pullrequestreview-5976711:

> @@ -6,4 +6,3 @@
>  /bin/
>  /obj/

-/*.xproj.user

This shouldn't be removed. *.xproj.user files are per-user settings, meaning yours may be different from mine.

---

In Controllers/HomeController.cshttps://github.com/Zoltu/bags-client/pull/58#pullrequestreview-5976711:

> -            ViewData["description"] = "What is your dream bag? Are you having trouble finding it? Let us help you!";
>   +
> -            return View("Index");
> -        }
>   +
> -        [Route("app/product/{productId}", Name = "product", Order = 0)]
> -        public IActionResult product(int productId)
> -        {
> -            ViewData["url"] = "Muhammad Nabeel";
> -            ViewData["type"] = "Managing Director";
> -            ViewData["title"] = "Principal Software Engineer / Technical Lead";
> -            ViewData["description"] = "Relliks Systems";
>   +
> -            return View("Index");
> -        }
> -    }

This project's code style is tab for indent, rather than spaces. The easiest way to address this is to install the EditorConfig plugin for Visual Studio, which will follow the indentation style rules found in .editorconfig automatically. Alternatively, you can change your Visual Studio settings to tabs, but you will have to switch it back and forth for every project you work on which can be a bit of a pain.

https://github.com/editorconfig/editorconfig-visualstudio#readme

---

In Startup.cshttps://github.com/Zoltu/bags-client/pull/58#pullrequestreview-5976711:

> -                applicationBuilder.UseExceptionHandler("/Home/Error");
> -            }
>   +
> -            applicationBuilder.UseApplicationInsightsRequestTelemetry();
> -                       applicationBuilder.UseApplicationInsightsExceptionTelemetry();
> -                       applicationBuilder.UseDefaultFiles();
> -                       applicationBuilder.UseStaticFiles();
>   +
> -            applicationBuilder.UseMvc(routes =>
> -            {
> -                routes.MapRoute(
> -                    name: "default",
> -                    template: "{controller=Home}/{action=Index}/{id?}");
> -            });
> -        }
> -       }

This file appears to have mixed tabs and spaces, see comment above about EditorConfig plugin.

---

In bags-client.slnhttps://github.com/Zoltu/bags-client/pull/58#pullrequestreview-5976711:

> +EndProject
> +Global
> -       GlobalSection(SolutionConfigurationPlatforms) = preSolution
> -               Debug|Any CPU = Debug|Any CPU
> -               Release|Any CPU = Release|Any CPU
> -       EndGlobalSection
> -       GlobalSection(ProjectConfigurationPlatforms) = postSolution
> -               {6BE8394A-B198-4B1B-80F5-8EB947F3C3F5}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
> -               {6BE8394A-B198-4B1B-80F5-8EB947F3C3F5}.Debug|Any CPU.Build.0 = Debug|Any CPU
> -               {6BE8394A-B198-4B1B-80F5-8EB947F3C3F5}.Release|Any CPU.ActiveCfg = Release|Any CPU
> -               {6BE8394A-B198-4B1B-80F5-8EB947F3C3F5}.Release|Any CPU.Build.0 = Release|Any CPU
> -       EndGlobalSection
> -       GlobalSection(SolutionProperties) = preSolution
> -               HideSolutionNode = FALSE
> -       EndGlobalSection
>   +EndGlobal

It looks like your settings also are messing with the line-endings in files. If you get the .editorconfig mentioned previously I believe it will do the right thing going forward. In the meantime, it doesn't appear there are any actual changes in this file, so you can just remove it from the change set.

---

In bags-client.xproj.userhttps://github.com/Zoltu/bags-client/pull/58#pullrequestreview-5976711:

> @@ -0,0 +1,6 @@
> +﻿<?xml version="1.0" encoding="utf-8"?>
> +<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
> -  <PropertyGroup>
> -    <ActiveDebugProfile>Kestrel</ActiveDebugProfile>
> -  </PropertyGroup>
>   +</Project>

As mentioned previously, this should not be committed to source control.

---

In client/scripts/templates/template.min.jshttps://github.com/Zoltu/bags-client/pull/58#pullrequestreview-5976711:

> @@ -0,0 +1,254 @@
> +(function() {

This file shouldn't be committed to source control. I'll add it to the .gitignore so it doesn't accidentally get re-added in a future PR.

---

In bags-client.slnhttps://github.com/Zoltu/bags-client/pull/58#pullrequestreview-5976711:

> +EndProject
> +Global
> -       GlobalSection(SolutionConfigurationPlatforms) = preSolution
> -               Debug|Any CPU = Debug|Any CPU
> -               Release|Any CPU = Release|Any CPU
> -       EndGlobalSection
> -       GlobalSection(ProjectConfigurationPlatforms) = postSolution
> -               {6BE8394A-B198-4B1B-80F5-8EB947F3C3F5}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
> -               {6BE8394A-B198-4B1B-80F5-8EB947F3C3F5}.Debug|Any CPU.Build.0 = Debug|Any CPU
> -               {6BE8394A-B198-4B1B-80F5-8EB947F3C3F5}.Release|Any CPU.ActiveCfg = Release|Any CPU
> -               {6BE8394A-B198-4B1B-80F5-8EB947F3C3F5}.Release|Any CPU.Build.0 = Release|Any CPU
> -       EndGlobalSection
> -       GlobalSection(SolutionProperties) = preSolution
> -               HideSolutionNode = FALSE
> -       EndGlobalSection
>   +EndGlobal

Actually, it appears this file may have had the wrong (or right depending on how you look at it) line endings previously... I'll investigate and figure out what the right answer is but in the meantime you should still remove this from the change set.

---

In project.jsonhttps://github.com/Zoltu/bags-client/pull/58#pullrequestreview-5976711:

> @@ -13,8 +14,14 @@
>                 "Microsoft.AspNetCore.Server.Kestrel": "1.0.1",
>                 "Microsoft.AspNetCore.StaticFiles": "1.0.0",
>                 "Microsoft.Extensions.Configuration.EnvironmentVariables": "1.0.0",
> -               "Microsoft.Extensions.Logging.Console": "1.0.0"
> -       },
> -    "Microsoft.Extensions.Logging.Console": "1.0.0",
> -    "Microsoft.AspNetCore.Mvc": "1.0.1",
> -    "Microsoft.AspNetCore.Routing": "1.0.1",
> -    "Microsoft.AspNetCore.Diagnostics": "1.0.0",
> -    "Microsoft.VisualStudio.Web.BrowserLink.Loader": "14.0.0",
> -    "Microsoft.AspNetCore.Razor.Tools": "1.0.0-preview2-final",
> -    "Microsoft.Extensions.Configuration.FileExtensions": "1.0.0",
> -    "Microsoft.Extensions.Configuration.Json": "1.0.0" },

As previously, the indentation in this file is off. This might have just been Visual Studio, it has a tendency to format this file wrong sometimes. Either way, please switch it to tabs.

---

In project.jsonhttps://github.com/Zoltu/bags-client/pull/58#pullrequestreview-5976711:

> @@ -13,8 +14,14 @@
>                 "Microsoft.AspNetCore.Server.Kestrel": "1.0.1",
>                 "Microsoft.AspNetCore.StaticFiles": "1.0.0",
>                 "Microsoft.Extensions.Configuration.EnvironmentVariables": "1.0.0",
> -               "Microsoft.Extensions.Logging.Console": "1.0.0"
> -       },
> -    "Microsoft.Extensions.Logging.Console": "1.0.0",
> -    "Microsoft.AspNetCore.Mvc": "1.0.1",
> -    "Microsoft.AspNetCore.Routing": "1.0.1",
> -    "Microsoft.AspNetCore.Diagnostics": "1.0.0",
> -    "Microsoft.VisualStudio.Web.BrowserLink.Loader": "14.0.0",
> -    "Microsoft.AspNetCore.Razor.Tools": "1.0.0-preview2-final",
> -    "Microsoft.Extensions.Configuration.FileExtensions": "1.0.0",
> -    "Microsoft.Extensions.Configuration.Json": "1.0.0" },

These two extensions shouldn't be necessary (Json and FileExtensions). Similarly, I don't think BrowserLink.Loader is necessary nor is Diagnostics. Recommend removing them and see if everything still works.

I'm not sure what Razor.Tools is for, I have a sneaking suspicion it isn't necessary either but I'm less confident. Can you try removing it and seeing if things still work?

---

In web.confighttps://github.com/Zoltu/bags-client/pull/58#pullrequestreview-5976711:

> @@ -0,0 +1,14 @@
> +﻿<?xml version="1.0" encoding="utf-8"?>
> +<configuration>
> +
> -  
>   +
> -  <system.webServer>
> -    <handlers>
> -      <add name="aspNetCore" path="*" verb="*" modules="AspNetCoreModule" resourceType="Unspecified"/>
> -    </handlers>
> -    <aspNetCore processPath="%LAUNCHER_PATH%" arguments="%LAUNCHER_ARGS%" stdoutLogEnabled="false" stdoutLogFile=".\logs\stdout" forwardWindowsAuthToken="false"/>
> -  </system.webServer>
>   +</configuration>

This file shouldn't be necessary since the app runs in Kestrel. I believe this is only necessary for IIS hosted applications, so it should be safe to delete.

---

In Startup.cshttps://github.com/Zoltu/bags-client/pull/58#pullrequestreview-5976711:

> -            else
> -            {
> -                applicationBuilder.UseExceptionHandler("/Home/Error");
> -            }
>   +
> -            applicationBuilder.UseApplicationInsightsRequestTelemetry();
> -                       applicationBuilder.UseApplicationInsightsExceptionTelemetry();
> -                       applicationBuilder.UseDefaultFiles();
> -                       applicationBuilder.UseStaticFiles();
>   +
> -            applicationBuilder.UseMvc(routes =>
> -            {
> -                routes.MapRoute(
> -                    name: "default",
> -                    template: "{controller=Home}/{action=Index}/{id?}");
> -            });

It appears that you are using attribute naming for routing, so I don't think this block is necessary anymore. In fact, as it is setup right now it appears to add a route that isn't actually (shouldn't be) a valid route. Can we remove this whole UseMvc(...) block and have everything still work?

---

In Startup.cshttps://github.com/Zoltu/bags-client/pull/58#pullrequestreview-5976711:

> -            //applicationBuilder.Use(async (context, next) =>
> -            //{
> -            // if (context.Request.Path.StartsWithSegments(new PathString("/app")))
> -            //         context.Request.Path = new PathString("/index.html");
> -            // await next();
> -            //});
>   +
> -            if (hostingEnvironment.IsDevelopment())
> -            {
> -                applicationBuilder.UseDeveloperExceptionPage();
> -                applicationBuilder.UseBrowserLink();
> -            }
> -            else
> -            {
> -                applicationBuilder.UseExceptionHandler("/Home/Error");
> -            }

All of this can be deleted I think. At the least, the exception handler and developer exception page. I have never been a fan of using browser link so I would recommend deleting that one as well for now. We may bring the BrowserLink stuff back later, but at this phase I would prefer to keep things trim and simple and since UI development isn't being done inside the VS project at the moment it is unnecessary.

---

In Startup.cshttps://github.com/Zoltu/bags-client/pull/58#pullrequestreview-5976711:

> -                    previous?.Invoke(context);
> -                };
> -            });
> -        }
>   +
> -               public void Configure(IApplicationBuilder applicationBuilder, IHostingEnvironment hostingEnvironment, ILoggerFactory loggerFactory)
> -               {
> -                       loggerFactory.AddConsole(minLevel: LogLevel.Warning);
>   +
> -            // custom middleware to re-route all paths starting with `/app` to `/index.html`
> -            //applicationBuilder.Use(async (context, next) =>
> -            //{
> -            // if (context.Request.Path.StartsWithSegments(new PathString("/app")))
> -            //         context.Request.Path = new PathString("/index.html");
> -            // await next();
> -            //});

Delete this code if it is unused. However, I suspect this code (or something similar is actually necessary to get all of the routes hooked up. At the moment it appears that the hooked up route is [Route("app", Name = "app", Order = 0)], which I'm not sure will catch everything inside of /app. We need the controller action to be called for every route that starts with /app, which includes /app/aboutus, /app/foo, etc. We may break some of those out into separate manual routes later but for now we want a single catch-all route that handles everything starting with /app.

One way to achieve this may be to leave this middleware in place (uncomment it) and make /index.html be the only route handled by the controller action. The middleware will trigger first, redirecting everything in /app to /index.html and then MVC will pick up that route and render it.

---

In Startup.cshttps://github.com/Zoltu/bags-client/pull/58#pullrequestreview-5976711:

> -                             .Build();
> -               }
>   +
> -               public void ConfigureServices(IServiceCollection services)
> -               {
> -                       services.AddApplicationInsightsTelemetry(_configuration);
> -            services.AddMvc();
>   +
> -            services.Configure<RazorViewEngineOptions>(options =>
> -            {
> -                var previous = options.CompilationCallback;
> -                options.CompilationCallback = (context) =>
> -                {
> -                    previous?.Invoke(context);
> -                };
> -            });

I think this can be removed since we don't have any razor compilation settings to set. At the moment, this code is just adding an unnecessary layer and confusing to the code.

I think you can delete the entire services.Configure<RazorViewEngineOptions>(...) block.

---

In Startup.cshttps://github.com/Zoltu/bags-client/pull/58#pullrequestreview-5976711:

> -                             .UseWebRoot("client")
> -                               .UseStartup<Startup>()
> -                               .UseUrls("http://*:80")
> -                               .Build()
> -                               .Run();
> -               }
>   +
> -               private IConfiguration _configuration;
>   +
> -               public Startup(IHostingEnvironment hostingEnvironment)
> -               {
> -                       _configuration = new ConfigurationBuilder()
> -                               .AddApplicationInsightsSettings(developerMode: hostingEnvironment.IsDevelopment())
> -                       .SetBasePath(hostingEnvironment.ContentRootPath)
> -                .AddJsonFile("appsettings.json", optional: true, reloadOnChange: true)
> -                .AddJsonFile($"appsettings.{hostingEnvironment.EnvironmentName}.json", optional: true)

This service only gets settings right now from environment variables, so we shouldn't be using JSON settings as it adds confusion to the code without benefit. At the moment, I believe this service actually has no settings in fact.

---

In Startup.cshttps://github.com/Zoltu/bags-client/pull/58#pullrequestreview-5976711:

> -                             .UseKestrel()
> -                               .UseContentRoot(Directory.GetCurrentDirectory())
> -                               .UseWebRoot("client")
> -                               .UseStartup<Startup>()
> -                               .UseUrls("http://*:80")
> -                               .Build()
> -                               .Run();
> -               }
>   +
> -               private IConfiguration _configuration;
>   +
> -               public Startup(IHostingEnvironment hostingEnvironment)
> -               {
> -                       _configuration = new ConfigurationBuilder()
> -                               .AddApplicationInsightsSettings(developerMode: hostingEnvironment.IsDevelopment())
> -                       .SetBasePath(hostingEnvironment.ContentRootPath)

SetBasePath can also be removed if we aren't doing JSON based service configuration.

---

In Controllers/HomeController.cshttps://github.com/Zoltu/bags-client/pull/58#pullrequestreview-5976711:

> +using System.Threading.Tasks;
> +using Microsoft.AspNetCore.Mvc;
> +
> +namespace Zoltu.Bags.Client.Controllers
> +{
> -    public class HomeController : Controller
> -    {
> -        public IActionResult Index()
> -        {
> -            ViewData["url"] = "https://bagcupid.com/";
> -            ViewData["type"] = "website";
> -            ViewData["title"] = "Bag Cupid";
> -            ViewData["description"] = "What is your dream bag? Are you having trouble finding it? Let us help you!";
>   +
> -            return View();
> -        }

I think we can delete this. The action that matters is the one below on line 22, this is just a duplicate.

---

In Controllers/HomeController.cshttps://github.com/Zoltu/bags-client/pull/58#pullrequestreview-5976711:

> +namespace Zoltu.Bags.Client.Controllers
> +{
> -    public class HomeController : Controller
> -    {
> -        public IActionResult Index()
> -        {
> -            ViewData["url"] = "https://bagcupid.com/";
> -            ViewData["type"] = "website";
> -            ViewData["title"] = "Bag Cupid";
> -            ViewData["description"] = "What is your dream bag? Are you having trouble finding it? Let us help you!";
>   +
> -            return View();
> -        }
>   +
> -        [Route("app", Name = "app", Order = 0)]
> -        public IActionResult app()

Style: Function names should start with a capital letter. App() in this case.

---

In Controllers/HomeController.cshttps://github.com/Zoltu/bags-client/pull/58#pullrequestreview-5976711:

> -            ViewData["title"] = "Bag Cupid";
> -            ViewData["description"] = "What is your dream bag? Are you having trouble finding it? Let us help you!";
>   +
> -            return View("Index");
> -        }
>   +
> -        [Route("app/product/{productId}", Name = "product", Order = 0)]
> -        public IActionResult product(int productId)
> -        {
> -            ViewData["url"] = "Muhammad Nabeel";
> -            ViewData["type"] = "Managing Director";
> -            ViewData["title"] = "Principal Software Engineer / Technical Lead";
> -            ViewData["description"] = "Relliks Systems";
>   +
> -            return View("Index");
> -        }

This method appears to be test code, so it shouldn't be included in the code. Please remove it from the change set.

---

In Controllers/HomeController.cshttps://github.com/Zoltu/bags-client/pull/58#pullrequestreview-5976711:

> -            ViewData["type"] = "website";
> -            ViewData["title"] = "Bag Cupid";
> -            ViewData["description"] = "What is your dream bag? Are you having trouble finding it? Let us help you!";
>   +
> -            return View();
> -        }
>   +
> -        [Route("app", Name = "app", Order = 0)]
> -        public IActionResult app()
> -        {
> -            ViewData["url"] = "https://bagcupid.com/";
> -            ViewData["type"] = "website";
> -            ViewData["title"] = "Bag Cupid";
> -            ViewData["description"] = "What is your dream bag? Are you having trouble finding it? Let us help you!";
>   +
> -            return View("Index");

The way this is setup right now, it is going to look for Views/Home/index.cshtml. I would like to figure out a way to make it so instead it gets its view from client/index.html instead, but parses it as a cshtml file. This would make it so my client developers don't have to change their workflow in order to mess with this file. This is the part that I don't know enough about Razor to do, and may require a little research to figure out.

---

In Controllers/HomeController.cshttps://github.com/Zoltu/bags-client/pull/58#pullrequestreview-5976711:

> @@ -0,0 +1,43 @@
> +﻿using System;
> +using System.Collections.Generic;
> +using System.Linq;
> +using System.Threading.Tasks;
> +using Microsoft.AspNetCore.Mvc;
> +
> +namespace Zoltu.Bags.Client.Controllers
> +{
> -    public class HomeController : Controller

Lets rename this to AppController.

.gitignore
inabeel: Didn't modified this file

Controllers/HomeController.cs
inabeel: HomeController is used to navigate default route to localhost. It works like Home link.
Unless necessary, Default controller shouldn't be renamed.

Controllers/HomeController.cs
inabeel: Index action in HomeController is to handle route localhost.
Use will be directly landing on this action and View Index.cshtml will be returned.

inabeel: Action method "app" is added to handle localhost/app path in the application.
in the client index.html page, when popup closes this path is shown as url.

And if User directly enters this url, he should be redirected to home page, so i added this action specifically for this purpose.

Controllers/HomeController.cs
inabeel: Changed.

Controllers/HomeController.cs
inabeel: It can return index.html but i feel that is not necessary.
Razor view hasn't anything to conflict with, and your client developers can change directly inside .cshtml file. 

Look at View inside Views/Home/index.chtml it has all same content with only meta changes, which needs to be rendered dynamically.

Controllers/HomeController.cs
inabeel: This Action method will be rendered to show product popup when rendered from the server. These details should be replaced with product details coming from Ajax call.

That is where I was asking to do another hit from server, to fill meta tags of the page.

Controllers/HomeController.cs
inabeel: I have installed this plugin and checked in some file. Please check if it works, or i will have to look into this in detail, plus we can discuss in detail during weekend.

Startup.cs
inabeel: Removed this..

Startup.cs
inabeel: Removed this

Startup.cs
inabeel: Removed this block and checked in Startup.cs

Startup.cs
inabeel: Default route in startup.cs will redirect user straight to Home controller Index Action to display your Index.Html file which is converted to View

Startup.cs
inabeel: Removed BrowserLink and this code from file

Startup.cs
inabeel: Default route in startup.cs will redirect user straight to Home controller Index Action to display your Index.Html file which is converted to View

Startup.cs
inabeel: Other relevant routing could be handled through Attribute Routing as done in Home Controller at this moment

bags-client.sln
inabeel: I have got that editor

bags-client.xproj.user
inabeel: I reverted it, seems didnt work

client/scripts/templates/template.min.js
inabeel: Ok

Controllers/HomeController.cs
MicahZoltu: The issue is that the client developers don't currently use Visual Studio to do their development and testing.  They just load index.html from where it is at into a browser, or host it with express (node) or any other web server.  Moving to `Views/Home/index.cshtml` will require all UI developers to now use Visual Studio for UI development work which is what I am trying to avoid.

Controllers/HomeController.cs
MicahZoltu: The way the client-side routing is setup, we won't be able to do real MVC based routing.  For example, this is a valid route: `/app/aboutus/product/14` and this is also a valid route that shows the same product: `/app/product/14`.  Because of this, we can't have a product route and a tags route, because the route can have both product and tags and price filter and aboutus.

.gitignore
MicahZoltu: It is possible your editor (Visual Studio) added it for you automatically.  You may need to do something to tell VS to not make this change.  At the least, revert the change in git so it doesn't show up in this pull request.

Controllers/HomeController.cs
MicahZoltu: See my comment below about not being able to use the MVC routing engine directly.  Because of this, the actions will end up needing to have some route parsing logic in them that define how they work, and at that point there isn't really any point in having a separate index route vs the "everything else route" since you will have to put all of the same logic into both of them.

Controllers/HomeController.cs
MicahZoltu: It seems that this file has its indentation fixed, but the line ending fix didn't take effect.  :/  I have made a change to the upstream repository (`Zoltu/bags-client`) that should make git do the right thing in the future, but in order to get that you will need to revert the change you made to `.gitignore` and then do a rebase on `origin/master`.  I can work with you on how to do that, or there are docs online for how to do that.  However, that won't resolve the line ending problem right now, only in the future.

The easiest solution is to install this extension: https://visualstudiogallery.msdn.microsoft.com/b2bbadd2-c337-43d7-9343-752ebbdd900f and right click in Visual Studio on the files you added/changed and unify the line endings to `Linux (LF)`.

Startup.cs
```diff
@@ -1,59 +1,61 @@
...
+			applicationBuilder.UseApplicationInsightsExceptionTelemetry();
+			applicationBuilder.UseDefaultFiles();
+			applicationBuilder.UseStaticFiles();
+
+			applicationBuilder.UseMvc(routes =>
+			{
+				routes.MapRoute(
+					name: "default",
+					template: "{controller=Home}/{action=Index}/{id?}");
+			});
```
MicahZoltu: Can we use Attribute Routing exclusively so we don't need this block here at all?  There are some Stack Overflow answers on how to set it up.  I would like to have routing all done the same way throughout the project, rather than having some routing done via this mechanism, and others done via attributes.

MicahZoltu: Also, `/Home/Index` should _not_ go to the index page.  It should load whatever static asset is at `/Home/Index`.  The only routes that should load the index page are:
- `/`
- `/app`
- `/app/*`

bags-client.sln
MicahZoltu: See my other comment about the line ending VS plugin to easily resolve the line ending problem.  I'm not sure why the `.editorconfig` isn't fixing the line endings for you on these files...

bags-client.xproj.user
MicahZoltu: There are a number of ways to resolve this problem.  One would be to delete the file, commit the delete, then revert the change you have to `.gitignore`.  Alternatively, you could do some git magic to reset your changes and create a new commit that fixes the issue, but that is more complicated.

client/scripts/templates/template.min.js
MicahZoltu: This file is still here, appears your attempt to delete it didn't work?

project.json
```diff
@@ -13,8 +14,11 @@
 		"Microsoft.AspNetCore.Server.Kestrel": "1.0.1",
 		"Microsoft.AspNetCore.StaticFiles": "1.0.0",
 		"Microsoft.Extensions.Configuration.EnvironmentVariables": "1.0.0",
-		"Microsoft.Extensions.Logging.Console": "1.0.0"
-	},
+    "Microsoft.Extensions.Logging.Console": "1.0.0",
+    "Microsoft.AspNetCore.Mvc": "1.0.1",
+    "Microsoft.AspNetCore.Routing": "1.0.1",
+    "Microsoft.AspNetCore.Razor.Tools": "1.0.0-preview2-final",
+    "Microsoft.Extensions.Configuration.Json": "1.0.0"	},
```
MicahZoltu: `Microsoft.Extensions.Configuration.Json` should be removed as it isn't being used.

project.json
```diff
@@ -1,9 +1,10 @@
...
 	"version": "1.0.0-*",
-	"buildOptions": {
-		"emitEntryPoint": true,
-		"debugType": "portable"
-	},
+  "buildOptions": {
+    "emitEntryPoint": true,
+    "debugType": "portable",
+    "preserveCompilationContext": true
+  },
```
MicahZoltu: Indentation is still not right for this file.  It may be be caused by Visual Studio editing the file on your behalf (for example when adding/removing NuGet packages.

MicahZoltu (changes requested):
To resolve the last of the styling problems, check out the https://visualstudiogallery.msdn.microsoft.com/b2bbadd2-c337-43d7-9343-752ebbdd900f extension and apply it to the files you have touched (including `project.lock.json` set to `Linux (LF)` line endings.  Future changes shouldn't have this problem as I have added a `.gitattributes` file to the repo that should avoid it in the future.

I think the biggest real piece of feedback is around the MVC routing stuff.  I am pretty confident (happy to be proven wrong) that we won't be able to use MVC's routing engine to do more than assign a couple catch-all routes (I included a list in one of my comments).  Let me know if what I am saying in the comments doesn't make sense, the way client-side routing occurs is non traditional which is why we can't leverage MVC's built in routing system.

Lastly, see my comments on the `client/index.html` vs `Views/Home/Index.cshtml`.  The issue is that I want the project to still be usable _without_ Visual Studio which means the file needs to be named `index.html` and live in `/client` in order for the UI developers to not have to change workflow/tooling.  We can discuss more if this doesn't make sense still.

.gitignore
inabeel: Reverted..

Controllers/HomeController.cs
inabeel: Installed plugin, and checked in files after unified.

Controllers/HomeController.cs
inabeel: MVC routing engine is used for calling all default routes index.
Some exception rules could be handled through Attribute Routing.

But if you have multiple routes to be used, under one pattern, then it is mentioned in Startup.cs to be used as one of default routes.

Controllers/HomeController.cs
inabeel: We need Razor engine just for Meta Tags. And meta tags wont effect any development changes.

Controllers/HomeController.cs
inabeel: At this moment, client side and server side routing working side by side.
On server when this url is provided [Route("app/product/{productId}", Name = "product", Order = 0)], we are returning our default index.cshtml.

When your index loads, it calls for client side routing against url to show popup.

bags-client.sln
inabeel: couldn't fix it.. need your time on this.,

client/scripts/templates/template.min.js
inabeel: yup i deleted, but seems it revives somehow

client/index.html
```diff
@@ -1,418 +1,418 @@
...
-
-		(function() {
-			var ur = document.createElement('script'); ur.type = 'text/javascript'; ur.async = true;
-			ur.src = ('https:' == document.location.protocol ? 'https://cdn.userreport.com/userreport.js' : 'http://cdn.userreport.com/userreport.js');
-			var s = document.getElementsByTagName('script')[0]; s.parentNode.insertBefore(ur, s);
-		})();
-	</script>
-</body>
-</html>
+﻿<!DOCTYPE html>
```
MicahZoltu: Same problem as before, it appears line endings were changed in this file, which makes the diff difficult to view.  You can fix this in VS by right clicking the file and choosing "Unify Line Endings" and then choosing "Linux (LF)".

Startup.cs
```diff
@@ -36,24 +38,33 @@ public Startup(IHostingEnvironment hostingEnvironment)
 		public void ConfigureServices(IServiceCollection services)
 		{
 			services.AddApplicationInsightsTelemetry(_configuration);
-			services.AddMvcCore();
+			services.AddMvc();
+
+			services.Configure<RazorViewEngineOptions>(options =>
+			{
+				options.ViewLocationExpanders.Add(new HTMLLocationRemapper());
+			});
```
MicahZoltu: Style: Since this is so simple, can you make it one-line without the curly braces?

```
services.Configure<RazorViewEngineOptions>(options => options.ViewLocationExpanders.Add(new HTMLLocationRemapper()));
```

client/index.html
```diff
@@ -1,418 +1,418 @@
...
+		(function() {
+			var ur = document.createElement('script'); ur.type = 'text/javascript'; ur.async = true;
+			ur.src = ('https:' == document.location.protocol ? 'https://cdn.userreport.com/userreport.js' : 'http://cdn.userreport.com/userreport.js');
+			var s = document.getElementsByTagName('script')[0]; s.parentNode.insertBefore(ur, s);
+		})();
+	</script>
+</body>
+</html>
+Contact GitHub API Training Shop Blog About
+© 2016 GitHub, Inc. Terms Privacy Security Status Help
```
MicahZoltu: This footer needs to be deleted, not sure how it got added (probably by a tool of some kind).

client/index.html
```diff
@@ -1,418 +1,418 @@
...
+		};
+	</script>
+
+	<script async src='https://www.google-analytics.com/analytics.js'></script>
+
+	<script async src='/vendors/google-analytics/autotrack.js'></script>
+	<!-- End Google Analytics -->
+	<!-- Bing Ads -->
+	<script>
+(function(w,d,t,r,u){var f,n,i;w[u]=w[u]||[],f=function(){var o={ti:"5438211"};o.q=w[u],w[u]=new UET(o),w[u].push("pageLoad")},n=d.createElement(t),n.src=r,n.async=1,n.onload=n.onreadystatechange=function(){var s=this.readyState;s&&s!=="loaded"&&s!=="complete"||(f(),n.onload=n.onreadystatechange=null)},i=d.getElementsByTagName(t)[0],i.parentNode.insertBefore(n,i)})(window,document,"script","//bat.bing.com/bat.js","uetq");
```
MicahZoltu: Having this on a newline is good, but it should be indented appropriately:

```
    <script>
        (function(w,d,t,r,u){var f,n,i;w[u]=w[u]||[],f=function(){var o={ti:"5438211"};o.q=w[u],w[u]=new UET(o),w[u].push("pageLoad")},n=d.createElement(t),n.src=r,n.async=1,n.onload=n.onreadystatechange=function(){var s=this.readyState;s&&s!=="loaded"&&s!=="complete"||(f(),n.onload=n.onreadystatechange=null)},i=d.getElementsByTagName(t)[0],i.parentNode.insertBefore(n,i)})(window,document,"script","//bat.bing.com/bat.js","uetq");        +(function(w,d,t,r,u){var f,n,i;w[u]=w[u]||[],f=function(){var o={ti:"5438211"};o.q=w[u],w[u]=new UET(o),w[u].push("pageLoad")},n=d.createElement(t),n.src=r,n.async=1,n.onload=n.onreadystatechange=function(){var s=this.readyState;s&&s!=="loaded"&&s!=="complete"||(f(),n.onload=n.onreadystatechange=null)},i=d.getElementsByTagName(t)[0],i.parentNode.insertBefore(n,i)})(window,document,"script","//bat.bing.com/bat.js","uetq");
    </script>
```

Controllers/HomeController.cs
```diff
@@ -0,0 +1,34 @@
...
+		[Route("app/product/{productId}", Name = "product", Order = 0)]
+		public IActionResult Product(int productId)
+		{
+			ViewData["url"] = "Muhammad Nabeel";
+			ViewData["type"] = "Managing Director";
+			ViewData["title"] = "Principal Software Engineer / Technical Lead";
+			ViewData["description"] = "Relliks Systems";
+
+			return View("Index");
+		}
```
MicahZoltu: This action should be removed, only the `App()` action should remain for now.

Controllers/HomeController.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Zoltu.Bags.Client.Controllers
+{
+	public class HomeController : Controller
+	{
+		[Route("", Name ="Home", Order = 0)]
+		[Route("app", Name = "app", Order = 0)]
+		public IActionResult App()
```
MicahZoltu: There is one route missing here, the `[Route("app/{*path}"]` route.  It should have an `Order` that is lower than the `"app"` route so it is evaluated only if `app` doesn't match.

This will require adding a parameter to the `App()` method for this with a default value of null:

```
public IActionResult App(String path = null)
```

Controllers/HomeController.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Zoltu.Bags.Client.Controllers
+{
+	public class HomeController : Controller
+	{
+		[Route("", Name ="Home", Order = 0)]
+		[Route("app", Name = "app", Order = 0)]
```
MicahZoltu: What is the `Name` on a route used for?  Can we remove it?

MicahZoltu: I ask because at the moment, these names are not meaningful to me and add complexity to the code because when I read them, I try to figure out how they are used and can't find anything.

MicahZoltu: I wonder if we should make the `Order` of the `app` route `1` just for consistency.  At the moment it has the same `Order` as the base route.  I suspect no problems will be caused with shared order, but I think being explicit will help make it more clear to the reader which routes should be evaluated first.

client/index.html
```diff
@@ -1,418 +1,418 @@
...
+			}
+		}
+		/* ]]> */
+	</script>
+	<script type="text/javascript" src="//www.googleadservices.com/pagead/conversion_async.js"></script>
+	<!-- End Google Ads Conversion Tracking -->
+	<meta property="og:url" content="@ViewBag.url">
+	<meta property="og:type" content="@ViewBag.type">
+	<meta property="og:title" content="@ViewBag.title">
+	<meta property="og:description" content="@ViewBag.descrition">
```
MicahZoltu: Typo here that is making this not work.

Controllers/HomeController.cs
```diff
@@ -0,0 +1,34 @@
...
+		public IActionResult App()
+		{
+			ViewData["url"] = "https://bagcupid.com/";
+			ViewData["type"] = "website";
+			ViewData["title"] = "Bag Cupid";
+			ViewData["description"] = "What is your dream bag? Are you having trouble finding it? Let us help you!";
+
+			return View("Index");
```
MicahZoltu: This needs to be lower case index because the server runs on Linux which is a case-sensitive operating system, so index.html and Index.html are not the same thing, which results in an exception being thrown here.

MicahZoltu (changes requested):
Along with the inline comments, one more change is necessary to make this work on the production site which is to add a line `COPY Controllers/ /app/Controllers/` to the `Dockerfile` after the line `COPY Startup.cs /app/`.  Without this, the newly added Controller won't make it into the docker image, and therefore won't be compiled in or deployed.

client/index.html
inabeel: Done

client/index.html
inabeel: Yup.. Added tab to indent

client/index.html
inabeel: Updated

client/index.html
inabeel: Footer removed.

Startup.cs
inabeel: Yup, done as per instruction

Controllers/HomeController.cs
inabeel: Name and Order in the route are parameters of Routing attribute. Name is used to name the route to identify each route uniquely. Order is given to define precedence of routes over other.
Since there is no any other routing, we can remove these.

Removed both attributes.

Controllers/HomeController.cs
inabeel: Following routes has been added on single action now. 

```
    [Route("")] // Default localhost
    [Route("app")] // to handle localhost/app
    [Route("app/{path}")] // to handle localhost/app/{anything}
    [Route("app/product/{productId}")] // to handle product pages
```

Controllers/HomeController.cs
inabeel: file and its name in action are now lower case

Controllers/HomeController.cs
inabeel: Removed, and all routing is added to single action

inabeel:
All changes has been performed, please verify and acknowledge.
Thank you,.

Controllers/HomeController.cs
```diff
@@ -0,0 +1,25 @@
...
+using Microsoft.AspNetCore.Mvc;
+
+namespace Zoltu.Bags.Client.Controllers
+{
+	public class HomeController : Controller
+	{
+		[Route("")]
+		[Route("app")]
+		[Route("app/{path}")]
+		[Route("app/product/{productId}")]
```
MicahZoltu: This route (`app/product/{productId}`) is already accounted for in the previous route (`app/{*path}`).

Controllers/HomeController.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+
+namespace Zoltu.Bags.Client.Controllers
+{
+	public class HomeController : Controller
+	{
+		[Route("")]
+		[Route("app")]
+		[Route("app/{path}")]
```
MicahZoltu: This needs to be a catch-all route: `app/{*path}`

There are a ton of different things that can go after the `/app/`, product is only one small portion of it.  The logic for parsing out the path will eventually occur inside the `App` action.

Controllers/HomeController.cs
```diff
@@ -0,0 +1,25 @@
...
+		[Route("app/product/{productId}")]
+		public IActionResult App(int productId = 0, string path = null)
```
MicahZoltu: Once you remove the route with `productId` in it, this parameter (`productId`) can then be removed.

MicahZoltu: Style: Use capital `S` `String`.

The reason for this is to get consistent syntax highlighting of class/struct names.  `String` in this case is interpreted as a class by the syntax highlighter, but `string` is interpreted as a keyword.  Also, all other classes start with a capital letter in all of these files, so having `String` be a special case leads to confusion.

Startup.cs
```diff
@@ -36,24 +38,30 @@ public Startup(IHostingEnvironment hostingEnvironment)
...
+
+	public class HTMLLocationRemapper : IViewLocationExpander
+	{
+		public IEnumerable<string> ExpandViewLocations(ViewLocationExpanderContext context, IEnumerable<string> viewLocations)
+		{
+			return viewLocations.Select(f => f.Replace("/Views/{1}/{0}.cshtml", "/client/{0}.html"));
+		}
+
+		public void PopulateValues(ViewLocationExpanderContext context) { }
+	}
```
MicahZoltu, line 66: This file once again has inconsistent line endings.  :/

If you catch me online over the weekend, I can show you some tools that will make it much easier to catch and deal with these issues.  Unfortunately, the Visual Studio git integration doesn't provide very good tooling for noticing these sorts of things and dealing with them.

MicahZoltu (changes requested):
The two big remaining issues are the missing line in the Dockerfile (`COPY Controllers/ /app/Controllers/` needs to be added after `COPY Startup.cs /app/`) and the comments I had on the routing.

Try to catch me online this weekend and I can show you how to test with Docker, and if you are interested, I can show you some tools that make avoiding a lot of these issues much easier.

PR #61, opened 2016-11-12
Getting Product details to fill meta tags

inabeel:


Controllers/HomeController.cs
```diff
@@ -3,21 +3,90 @@
...
+
+			if (match.Success)
+			{
+				return match.Groups.Count == 1 ? match.Value : match.Groups[1].Value;
+			}
+
+			return String.Empty;
+		}
+
+		static dynamic CallAPI(String path)
```
MicahZoltu: Style: Lowercase acronyms like `CallApi`

Controllers/HomeController.cs
```diff
@@ -3,21 +3,90 @@
...
+			String description = "What is your dream bag? Are you having trouble finding it? Let us help you!";
+			String image = "https://bagcupid.com/img/logo/bagcupid.png";
+
+			if (!String.IsNullOrEmpty(path))
+			{
+				String ProductId = MatchRegexAndReturnValue(path, @"product\/(\d+)");
+				String TagId = MatchRegexAndReturnValue(path, @"tags\/(.*?)\/*?$");
+				String MinPrice = MatchRegexAndReturnValue(path, @"minprice\/(\d+)");
+				String MaxPrice = MatchRegexAndReturnValue(path, @"maxprice\/(\d+)");
+				String aboutUs = MatchRegexAndReturnValue(path, @"aboutus");
```
MicahZoltu: Performance: Precompile the regular expressions in private static member variables.  This can have a huge impact on performance in some situations.

Controllers/HomeController.cs
```diff
@@ -3,21 +3,90 @@
...
+			ViewData["title"] = title;
+			ViewData["description"] = description;
+			ViewData["image"] = image;
 
 			return View("index");
 		}
+
+		public String MatchRegexAndReturnValue(String input, String pattern)
+		{
+			var match = Regex.Match(input, pattern);
```
MicahZoltu: Bug: These should all be case-insensitive matches, I believe there is a parameter that you can pass to `Regex.Match` to do that.

Controllers/HomeController.cs
```diff
@@ -3,21 +3,90 @@
...
 	public class HomeController : Controller
 	{
+		// Best practices
+		// https://www.asp.net/web-api/overview/advanced/calling-a-web-api-from-a-net-client
+		static HttpClient client = new HttpClient();
+
 		[Route("")]
 		[Route("app/{*path}")]
-		public IActionResult App(string path = null)
+		public IActionResult App(String path = null)
```
MicahZoltu: Performance: Now that we are calling an external service, this call stack should be changed to an async one.  This method and `CallApi` should be like `public async Task<IActionResult> App(String path = null)` and when you call one of them, call like `await CallApi(...)`

Controllers/HomeController.cs
```diff
@@ -3,21 +3,90 @@
...
+
+			var task = client.GetAsync(path);
+			task.Wait();
+
+			HttpResponseMessage response = task.Result;
+			dynamic obj = null;
+			if (response.IsSuccessStatusCode)
+			{
+				var contentTask = response.Content.ReadAsStringAsync();
+				contentTask.Wait();
```
MicahZoltu: Performance: Once you make this method and the action async, these two lines should change to:
`var content = await response.Content.ReadAsStringAsync();`

Controllers/HomeController.cs
```diff
@@ -3,21 +3,90 @@
...
+		static dynamic CallAPI(String path)
+		{
+			client.BaseAddress = new Uri("https://bags-api.zoltu.com/api/");
+			client.DefaultRequestHeaders.Accept.Clear();
+			client.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+
+			var task = client.GetAsync(path);
+			task.Wait();
+
+			HttpResponseMessage response = task.Result;
```
MicahZoltu: Performance: Once you make this method and the action async, these three lines should change to:
`var response = await client.GetAsync(path);`

Controllers/HomeController.cs
```diff
@@ -3,21 +3,90 @@
...
+			}
+
+			return String.Empty;
+		}
+
+		static dynamic CallAPI(String path)
+		{
+			client.BaseAddress = new Uri("https://bags-api.zoltu.com/api/");
+			client.DefaultRequestHeaders.Accept.Clear();
+			client.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
```
MicahZoltu: Performance: These should be set in a static constructor rather than here since they are always the same and we are re-using the client.

Controllers/HomeController.cs
```diff
@@ -3,21 +3,90 @@
...
+using System.Net.Http;
+using System.Net.Http.Headers;
+using Newtonsoft.Json;
 
 namespace Zoltu.Bags.Client.Controllers
 {
 	public class HomeController : Controller
 	{
+		// Best practices
+		// https://www.asp.net/web-api/overview/advanced/calling-a-web-api-from-a-net-client
```
MicahZoltu: Style: Thanks for the comment with the link, very helpful!  Can you put it all on one line though?
`// Best Practices: https://www.asp.net/web-api/overview/advanced/calling-a-web-api-from-a-net-client`

Controllers/HomeController.cs
MicahZoltu: Style: These can all be `var`s rather than `String`.

Controllers/HomeController.cs
```diff
@@ -3,21 +3,90 @@
...
+			{
+				String ProductId = MatchRegexAndReturnValue(path, @"product\/(\d+)");
+				String TagId = MatchRegexAndReturnValue(path, @"tags\/(.*?)\/*?$");
+				String MinPrice = MatchRegexAndReturnValue(path, @"minprice\/(\d+)");
+				String MaxPrice = MatchRegexAndReturnValue(path, @"maxprice\/(\d+)");
+				String aboutUs = MatchRegexAndReturnValue(path, @"aboutus");
+
+				if (string.IsNullOrEmpty(ProductId))
+				{
+					dynamic product = CallAPI("products/" + ProductId);
```
MicahZoltu: Style: This can be `var`.

Controllers/HomeController.cs
```diff
@@ -3,21 +3,90 @@
...
+
+			if (!String.IsNullOrEmpty(path))
+			{
+				String ProductId = MatchRegexAndReturnValue(path, @"product\/(\d+)");
+				String TagId = MatchRegexAndReturnValue(path, @"tags\/(.*?)\/*?$");
+				String MinPrice = MatchRegexAndReturnValue(path, @"minprice\/(\d+)");
+				String MaxPrice = MatchRegexAndReturnValue(path, @"maxprice\/(\d+)");
+				String aboutUs = MatchRegexAndReturnValue(path, @"aboutus");
+
+				if (string.IsNullOrEmpty(ProductId))
```
MicahZoltu: Style: Use upper case `String` here.

Controllers/HomeController.cs
```diff
@@ -3,21 +3,90 @@
...
+			client.BaseAddress = new Uri("https://bags-api.zoltu.com/api/");
+			client.DefaultRequestHeaders.Accept.Clear();
+			client.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+
+			var task = client.GetAsync(path);
+			task.Wait();
+
+			HttpResponseMessage response = task.Result;
+			dynamic obj = null;
+			if (response.IsSuccessStatusCode)
```
MicahZoltu: Style: Invert this if statement:

```
if (!response.IsSuccessStatusCode)
    return null;

var content = response.Content.ReadAsStringAsync();
return JsonConvert.DeserializeObject(content);
```

This will reduce the complexity of this function significantly, making it less prone to bugs getting introduced later.

Controllers/HomeController.cs
```diff
@@ -3,21 +3,90 @@
...
 		}
+
+		public String MatchRegexAndReturnValue(String input, String pattern)
+		{
+			var match = Regex.Match(input, pattern);
+
+			if (match.Success)
+			{
+				return match.Groups.Count == 1 ? match.Value : match.Groups[1].Value;
+			}
```
MicahZoltu: Style: For simple if statements like this, prefer no curly braces:

```
if (match.Success)
    return match.Groups.Count == 1 ? match.Value : match.Groups[1].Value;
```

Controllers/HomeController.cs
```diff
@@ -3,21 +3,90 @@
...
 			return View("index");
 		}
+
+		public String MatchRegexAndReturnValue(String input, String pattern)
+		{
+			var match = Regex.Match(input, pattern);
+
+			if (match.Success)
+			{
+				return match.Groups.Count == 1 ? match.Value : match.Groups[1].Value;
```
MicahZoltu: Possible Bug: Is it possible for `match.Success` to be true but `match.Groups.Count == 0`?  If so, this will throw an out of bounds exception when you try to index into `Groups[1]`.

Controllers/HomeController.cs
MicahZoltu: Bug: Isn't this supposed to be `if (!String.IsNullOrEmpty(productId))`?  At the moment, the following code block will only be executed if there is no `productId`, which I don't believe is correct (it should only be executed if there _is_ a product ID).

Controllers/HomeController.cs
```diff
@@ -3,21 +3,90 @@
...
+
+				if (string.IsNullOrEmpty(ProductId))
+				{
+					dynamic product = CallAPI("products/" + ProductId);
+
+					url = "https://bagcupid.com/" + path;
+					type = "website";
+					title = "Bag Cupid";
+					description = "What is your dream bag? Are you having trouble finding it? Let us help you!";
+					image = product.images[0].small;
```
MicahZoltu: Resiliency:  We should do some error checking here to make sure the images array has at least one item in it.

Controllers/HomeController.cs
MicahZoltu: Style: local variables should have lower case first letters like `productId` and `tagId`.

MicahZoltu (changes requested):
I tried to prefix all comments with the type of feedback it is, Style, Performance or Bug to help give an idea of why I find each of the things important, and to also weigh how important each is.

Also, I realize that now that this code has actual business logic, we should write some Unit Tests for it.  I'll setup the infrastructure for Unit Testing separately, and we can add them in a future PR.

Controllers/HomeController.cs
inabeel: Done, Made it on single line

Controllers/HomeController.cs
inabeel: Ok, i will fix it

Controllers/HomeController.cs
inabeel: I made it Static to get rid of disposing them at the end. I can write that in constructor as static member. So it wont reinitialize again and again.

inabeel: Normally as best practice its norm to use Type name instead of var. I will update it as per requirement,.
Function parameters are in lower first case Camel, and rest local variable should be Pascal in general. I will update it though.

Controllers/HomeController.cs
inabeel: Agree.. fixed

Controllers/HomeController.cs
MicahZoltu: I believe full type names is the legacy standard.  `var` was introduced fairly recently into the C# language and is the new general recommendation.  Many older projects however use full type names everywhere and being consistent within a project is more important than anything else, which is why often projects will prefer full type names.

Controllers/HomeController.cs
inabeel: fixed.

Controllers/HomeController.cs
MicahZoltu: The coding standard for this project (which I believe aligns with the .NET framework standard) is Pascal for everything except parameters and local variables.  Interestingly, I went to look this up and it appears that MS forgot to include local variables in their style guidelines.  :D  https://msdn.microsoft.com/en-us/library/ms229043(v=vs.110).aspx  That being said, I don't follow all of their guidelines so don't take that as a source of truth.  :)  As a point of reference, here is .NET Core Framework source code, which uses camelCase local variables: https://github.com/dotnet/corefx/blob/master/src/Microsoft.CSharp/src/Microsoft/CSharp/RuntimeBinder/Binder.cs#L35

Controllers/HomeController.cs
inabeel: There is question: Which image you want to choose to display? I picked small just for completeness purpose.

MicahZoltu: Good question.  We should probably use the large image for this purpose.  Most crawlers/bots will downsample it to the size they want.

On that topic, we should choose the image with the smallest `priority`.  If they are all equal we should choose the first one.

Controllers/HomeController.cs
inabeel: Fixed

Controllers/HomeController.cs
inabeel: In this line i wrote it to handle "about"
it wont b 0 any time if it matches
if it is 0 then it means it wont succeed, thus wont come in this block

Controllers/HomeController.cs
inabeel: i normally add them for cleanliness. I will take care of this now.
Fixed

Controllers/HomeController.cs
inabeel: Fixed

Controllers/HomeController.cs
inabeel: Fixed

Controllers/HomeController.cs
inabeel: Fixed

Controllers/HomeController.cs
inabeel: Fixed

Controllers/HomeController.cs
inabeel: Fixed

Controllers/HomeController.cs
inabeel: Great effort Thanks :)

Controllers/HomeController.cs
inabeel: Fixed

inabeel (commented):
Changes has been done.

Controllers/HomeController.cs
```diff
@@ -3,21 +3,88 @@
...
 	public class HomeController : Controller
 	{
+		private static Regex productRegx = new Regex(@"product\/(\d+)", RegexOptions.IgnoreCase);
+		private static Regex tagRegx = new Regex(@"tags\/(.*?)\/*?$", RegexOptions.IgnoreCase);
+		private static Regex minPriceRegx = new Regex(@"minprice\/(\d+)", RegexOptions.IgnoreCase);
+		private static Regex maxPriceRegx = new Regex(@"maxprice\/(\d+)", RegexOptions.IgnoreCase);
+		private static Regex aboutRegx = new Regex(@"aboutus", RegexOptions.IgnoreCase);
+		private static HttpClient client = null;
+
+		public HomeController()
```
MicahZoltu: Performance: This should be a static constructor so it is only executed once:

```
static HomeController()
{
    ...
}
```

Controllers/HomeController.cs
```diff
@@ -3,21 +3,88 @@
...
+		{
+			// Best practices: https://www.asp.net/web-api/overview/advanced/calling-a-web-api-from-a-net-client
+			client = new HttpClient();
```
MicahZoltu: Style: Inline this with the declaration:

```
private static readonly HttpClient client = new HttpClient()
```

Controllers/HomeController.cs
```diff
@@ -3,21 +3,88 @@
...
 namespace Zoltu.Bags.Client.Controllers
 {
 	public class HomeController : Controller
 	{
+		private static Regex productRegx = new Regex(@"product\/(\d+)", RegexOptions.IgnoreCase);
+		private static Regex tagRegx = new Regex(@"tags\/(.*?)\/*?$", RegexOptions.IgnoreCase);
+		private static Regex minPriceRegx = new Regex(@"minprice\/(\d+)", RegexOptions.IgnoreCase);
+		private static Regex maxPriceRegx = new Regex(@"maxprice\/(\d+)", RegexOptions.IgnoreCase);
+		private static Regex aboutRegx = new Regex(@"aboutus", RegexOptions.IgnoreCase);
+		private static HttpClient client = null;
```
MicahZoltu: Safety: All of these should be `readonly` to make it clear to the reader that they are set once here and never again.  It also helps prevent subtle bugs from getting added later.

Controllers/HomeController.cs
```diff
@@ -3,21 +3,88 @@
...
 
 namespace Zoltu.Bags.Client.Controllers
 {
 	public class HomeController : Controller
 	{
+		private static Regex productRegx = new Regex(@"product\/(\d+)", RegexOptions.IgnoreCase);
+		private static Regex tagRegx = new Regex(@"tags\/(.*?)\/*?$", RegexOptions.IgnoreCase);
+		private static Regex minPriceRegx = new Regex(@"minprice\/(\d+)", RegexOptions.IgnoreCase);
+		private static Regex maxPriceRegx = new Regex(@"maxprice\/(\d+)", RegexOptions.IgnoreCase);
+		private static Regex aboutRegx = new Regex(@"aboutus", RegexOptions.IgnoreCase);
```
MicahZoltu: Performance: These should all be compiled regular expressions.  https://msdn.microsoft.com/en-us/library/8zbs0h2f(v=vs.110).aspx

```
private static readonly Regex aboutRegx = new Regex(@"aboutus", RegexOptions.IgnoreCase | RegexOptions.Compiled);
```

Controllers/HomeController.cs
```diff
@@ -3,21 +3,88 @@
...
+			var url = "https://bagcupid.com/";
+			var type = "website";
+			var title = "Bag Cupid";
+			var description = "What is your dream bag? Are you having trouble finding it? Let us help you!";
+			var image = "https://bagcupid.com/img/logo/bagcupid.png";
+
+			if (!String.IsNullOrEmpty(path))
+			{
+				var productMatch = productRegx.Match(path);
+				var productId = productMatch.Success ? productMatch.Groups.Count == 1 ? productMatch.Value : productMatch.Groups[1].Value : String.Empty;
```
MicahZoltu: Bug: I don't believe the logic here is quite right.  Given the regular expression, I don't believe it is possible to match it (as it is now) with a single group.  That being said, the code should be robust against future changes and not set the productId if there is none.  I think the logic should be something like this in the end:

```
if (!productMatch.Success)
    productId = null
else if (productMatch.Groups.Count != 2)
    productId = null
else
    productId = productMatch.Groups[1].Value
```

This could probably be simplified a bit, but I wanted to be explicit in this comment.

MicahZoltu: Style: Don't use magic values.  In this case, if we don't have a `productId`, then set `productId` to null rather than an empty string (magic value).

Controllers/HomeController.cs
```diff
@@ -3,21 +3,88 @@
...
+
+				var tagMatch = tagRegx.Match(path);
+				var tagId = tagMatch.Success ? tagMatch.Groups.Count == 1 ? tagMatch.Value : tagMatch.Groups[1].Value : String.Empty;
```
MicahZoltu: Bug: Mostly the same comment as above, though in this case it is possible to match and have a group count of 1.  The logic should be the same as described above though, only setting `tagMatch` to non-null if there is exactly 2 match groups.

Controllers/HomeController.cs
```diff
@@ -3,21 +3,88 @@
...
+
+				var minPriceMatch = minPriceRegx.Match(path);
+				var minPrice = minPriceMatch.Success ? minPriceMatch.Groups.Count == 1 ? minPriceMatch.Value : minPriceMatch.Groups[1].Value : String.Empty;
+
+				var maxPriceMatch = maxPriceRegx.Match(path);
+				var maxPrice = maxPriceMatch.Success ? maxPriceMatch.Groups.Count == 1 ? maxPriceMatch.Value : maxPriceMatch.Groups[1].Value : String.Empty;
+
+				var aboutUsMatch = aboutRegx.Match(path);
+				var aboutUs = aboutUsMatch.Success ? aboutUsMatch.Groups.Count == 1 ? aboutUsMatch.Value : aboutUsMatch.Groups[1].Value : String.Empty;
```
MicahZoltu: Bug: `aboutUs` doesn't have a match group, so it will never have a group count of anything other than 1.  Also, `aboutUs` should be a boolean that is simply set to `var aboutUs = aboutUsMatch.Success`.  The above two lines could also be simplified into one: `var aboutUs = aboutRegEx.Match(path).Success`.

Controllers/HomeController.cs
```diff
@@ -3,21 +3,88 @@
...
+				var productId = productMatch.Success ? productMatch.Groups.Count == 1 ? productMatch.Value : productMatch.Groups[1].Value : String.Empty;
+
+				var tagMatch = tagRegx.Match(path);
+				var tagId = tagMatch.Success ? tagMatch.Groups.Count == 1 ? tagMatch.Value : tagMatch.Groups[1].Value : String.Empty;
+
+				var minPriceMatch = minPriceRegx.Match(path);
+				var minPrice = minPriceMatch.Success ? minPriceMatch.Groups.Count == 1 ? minPriceMatch.Value : minPriceMatch.Groups[1].Value : String.Empty;
+
+				var maxPriceMatch = maxPriceRegx.Match(path);
+				var maxPrice = maxPriceMatch.Success ? maxPriceMatch.Groups.Count == 1 ? maxPriceMatch.Value : maxPriceMatch.Groups[1].Value : String.Empty;
```
MicahZoltu: Same comment as for `productId` for both `min` and `max` price.

Controllers/HomeController.cs
MicahZoltu: Style: `Regex` should be spelled out in the variable names to match the name of the class that they are an instance of.

Controllers/HomeController.cs
```diff
@@ -3,21 +3,88 @@
...
+				var aboutUsMatch = aboutRegx.Match(path);
+				var aboutUs = aboutUsMatch.Success ? aboutUsMatch.Groups.Count == 1 ? aboutUsMatch.Value : aboutUsMatch.Groups[1].Value : String.Empty;
+
+				if (!String.IsNullOrEmpty(productId))
+				{
+					var product = await CallApi("products/" + productId);
+
+					url = "https://bagcupid.com/" + path;
+					type = "website";
+					title = "Bag Cupid";
```
MicahZoltu: Style: `type` and `title` aren't changed here, so don't reset them as it makes it unclear what the intent is and creates an opportunity to introduce bugs if someone changes one but not the other.

Controllers/HomeController.cs
```diff
@@ -3,21 +3,88 @@
...
+				var minPriceMatch = minPriceRegx.Match(path);
+				var minPrice = minPriceMatch.Success ? minPriceMatch.Groups.Count == 1 ? minPriceMatch.Value : minPriceMatch.Groups[1].Value : String.Empty;
+
+				var maxPriceMatch = maxPriceRegx.Match(path);
+				var maxPrice = maxPriceMatch.Success ? maxPriceMatch.Groups.Count == 1 ? maxPriceMatch.Value : maxPriceMatch.Groups[1].Value : String.Empty;
+
+				var aboutUsMatch = aboutRegx.Match(path);
+				var aboutUs = aboutUsMatch.Success ? aboutUsMatch.Groups.Count == 1 ? aboutUsMatch.Value : aboutUsMatch.Groups[1].Value : String.Empty;
+
+				if (!String.IsNullOrEmpty(productId))
```
MicahZoltu: Style: As per the above, an empty String shouldn't be used as a magic value.  Instead, check for null only here.

MicahZoltu: Robustness: We should validate the productId before this, possible up by the `Regex.Match` code.  If `productId` is not a number then it isn't a valid `productId` and we shouldn't enter into this code block.  The easiest way to achieve this would be to attempt to parse it as a number with `UInt64.TryParse(...)`.  Even better would be to create an extension method named `String.TryParseUInt64()` that returned null if the string didn't parse.  This would simplify the code significantly over using C#'s built-in `TryParse` method.

Controllers/HomeController.cs
```diff
@@ -3,21 +3,88 @@
...
+				{
+					var product = await CallApi("products/" + productId);
+
+					url = "https://bagcupid.com/" + path;
+					type = "website";
+					title = "Bag Cupid";
+					description = "What is your dream bag? Are you having trouble finding it? Let us help you!";
+					image = (product != null && product.images.Count > 0 && product.images[0].large != null) ? product.images[0].large : "";
```
MicahZoltu: Bug: This should return the highest priority image.  Each of the images should have a `Priority` field on it and the lowest number is the right one.  If multiple images have the same number, choose the first.

Controllers/HomeController.cs
```diff
@@ -3,21 +3,88 @@
...
-			ViewData["type"] = "website";
-			ViewData["title"] = "Bag Cupid";
-			ViewData["description"] = "What is your dream bag? Are you having trouble finding it? Let us help you!";
+			var url = "https://bagcupid.com/";
+			var type = "website";
+			var title = "Bag Cupid";
+			var description = "What is your dream bag? Are you having trouble finding it? Let us help you!";
+			var image = "https://bagcupid.com/img/logo/bagcupid.png";
+
+			if (!String.IsNullOrEmpty(path))
```
MicahZoltu, line 43: Style: This entire `if` block can be avoided by doing `path = path ?? ""`.  `productRegex.Match(path)` would then never throw an exception for bad input, and instead it simply wouldn't match anything.  In this case, I think the performance cost of executing the regular expressions are worth the reduction in code complexity (all branching increases code complexity).

Controllers/HomeController.cs
```diff
@@ -3,21 +3,88 @@
...
+using System.Text.RegularExpressions;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using Newtonsoft.Json;
 
 namespace Zoltu.Bags.Client.Controllers
 {
 	public class HomeController : Controller
 	{
+		private static Regex productRegx = new Regex(@"product\/(\d+)", RegexOptions.IgnoreCase);
```
MicahZoltu: Style: We could probably improve the readability of the code later on as well as reduce the chance for future bugs by using named captures.

```
var regex = new Regex(@"product\/(?<productId>\d+)");
var productId = regex.Match(path).Groups["productId"].Value.TryParseUInt64();

```

The `TryParseUInt64` extension method would need to be created (see other comment about this).

The same thing could be applied to all of the regular expressions that have match groups, though `tags` just returns a string so that wouldn't get run through the `TryParseUInt64` bit.

Controllers/HomeController.cs
MicahZoltu: The same goes for `description`.

Controllers/HomeController.cs
```diff
@@ -3,21 +3,88 @@
...
+				var maxPrice = maxPriceMatch.Success ? maxPriceMatch.Groups.Count == 1 ? maxPriceMatch.Value : maxPriceMatch.Groups[1].Value : String.Empty;
+
+				var aboutUsMatch = aboutRegx.Match(path);
+				var aboutUs = aboutUsMatch.Success ? aboutUsMatch.Groups.Count == 1 ? aboutUsMatch.Value : aboutUsMatch.Groups[1].Value : String.Empty;
+
+				if (!String.IsNullOrEmpty(productId))
+				{
+					var product = await CallApi("products/" + productId);
+
+					url = "https://bagcupid.com/" + path;
```
MicahZoltu: Bug: This should be set to `$"https://bagcupid.com/app/{path}"`.

Controllers/HomeController.cs
```diff
@@ -3,21 +3,88 @@
...
+
+				var maxPriceMatch = maxPriceRegx.Match(path);
+				var maxPrice = maxPriceMatch.Success ? maxPriceMatch.Groups.Count == 1 ? maxPriceMatch.Value : maxPriceMatch.Groups[1].Value : String.Empty;
+
+				var aboutUsMatch = aboutRegx.Match(path);
+				var aboutUs = aboutUsMatch.Success ? aboutUsMatch.Groups.Count == 1 ? aboutUsMatch.Value : aboutUsMatch.Groups[1].Value : String.Empty;
+
+				if (!String.IsNullOrEmpty(productId))
+				{
+					var product = await CallApi("products/" + productId);
```
MicahZoltu: Style: Use string interpolation for string building: `$"products/{productId}"`

Controllers/HomeController.cs
```diff
@@ -3,21 +3,88 @@
...
 		}
+
+		static async Task<dynamic> CallApi(String path)
+		{
+			var response = await client.GetAsync(path);
+
+			if (!response.IsSuccessStatusCode)
+				return null;
+
+			var contentTask = await response.Content.ReadAsStringAsync();
```
MicahZoltu: Style: `contentTask` should be renamed to `content` as it is no longer a task anymore.

MicahZoltu (changes requested):
Looking good!  I have some more feedback none the less.  :)

Controllers/HomeController.cs
inabeel: Fixed.

Controllers/HomeController.cs
inabeel: Fixed..
Last one HttpClient cannot be readonly as it requires assignment

Controllers/HomeController.cs
inabeel: Adding Static Constructor wont help.
1- Readonly fields of regex cannot be initialized later, even in constructor.
2- If we move whole declaration then, fields will lose its scope.
3- Static constructor itself wont be effective, as declaring and initializing will keep a copy of initialized regex field in memory. which should be enough

As per https://www.dotnetperls.com/static
Static constructor. A static constructor initializes static fields. It runs at an indeterminate time before those fields are used. Static constructors on a type impose some overhead.

Controllers/HomeController.cs
inabeel: Client object require in whole class, to be used inside CallApi function. So we cant move it inside constructor.
Secondly, since we need to initialize it in controller, we can't make it readonly.
finally, we want to dispose http client after every call, so we cant keep as class object as initialized.
so ideal flow is to make it class level property, initialize on controller call, and dispose on api call, and repeat.

Controllers/HomeController.cs
inabeel: Fixed

Controllers/HomeController.cs
inabeel: fixed

Controllers/HomeController.cs
inabeel: fixed

Controllers/HomeController.cs
MicahZoltu: See my other comment on this, the `HttpClient` should be `readonly` as well and initialized once at class initialization, rather than initialized every time in the class constructor.

Controllers/HomeController.cs
MicahZoltu: > we want to dispose http client after every call

I don't believe this is the case based on the link you provided in the comment above: https://www.asp.net/web-api/overview/advanced/calling-a-web-api-from-a-net-client

> HttpClient is intended to be instantiated once and re-used throughout the life of an application

What I am recommending is this:

```
public class HomeController : Controller
{
    private static readonly HttpClient httpClient = new HttpClient();
    static HomeController
    {
        client.BaseAddress = new Uri("https://bags-api.zoltu.com/api/");
        client.DefaultRequestHeaders.Accept.Clear();
        client.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
    }
}
```

This example would result in one `HttpClient` for the entire application.  If you initialize the `HttpClient` in the normal constructor (not a static one), then you end up with one `HttpClient` per request which is what the article you linked recommends _against_.  For the same reason, the `HttpClient` should not be disposed at the end of the request, it should live for the life of the application.

Controllers/HomeController.cs
MicahZoltu: See my other comment on this, the `HttpClient` is what should be setup in a static constructor.  The Regex items should be `readonly` and initialized inline.

MicahZoltu (changes requested):

Controllers/HomeController.cs
inabeel: Written extension method like this:

```
    public static UInt64? TryParseUInt64(this String value)
    {
        UInt64 id = 0;
        return UInt64.TryParse(value, out id) ? id : (UInt64?)null;
    }
```

Calling line:
    var productId = productRegex.Match(path).Groups[1].Value.TryParseUInt64();

and checking value using:
       if(productId.HasValue)

Controllers/HomeController.cs
inabeel: HttpClient also fixed.

Controllers/HomeController.cs
inabeel: Got it and fixed.

Controllers/HomeController.cs
inabeel: Agree got it.. Fixed

Controllers/HomeController.cs
inabeel: Not only Regex but also attempt to parse is another expensive operation.
Let me know what to do with it..

Controllers/HomeController.cs
inabeel: Changed this section to be like:

```
            var productId = productRegex.Match(path).Groups[1].Value?.TryParseUInt64();
            var tagId = tagRegex.Match(path).Groups[1].Value;
            var minPrice = minPriceRegex.Match(path).Groups[1].Value.TryParseUInt64();
            var maxPrice = maxPriceRegex.Match(path).Groups[1].Value.TryParseUInt64();
            var aboutUs = aboutRegex.Match(path).Success;
```

Controllers/HomeController.cs
inabeel: Exactly 2 match group?
I don't feel this might occur! as multiple tags come concatenated with underscore(_)

Controllers/HomeController.cs
inabeel: Fixed

Controllers/HomeController.cs
inabeel: Fixed.. Check and provide your feedback

Controllers/HomeController.cs
inabeel: fixed

Controllers/HomeController.cs
inabeel: fixed

Controllers/HomeController.cs
inabeel: Achieved it through possible means to my mind:

```
if (product?.images != null && product.images.Count > 0)
                    image = (string)(((JObject)product)["images"].Aggregate((minItem, nextItem) => (int)minItem["priority"] < (int)nextItem["priority"] ? minItem : nextItem))["large"];
```

inabeel:
All changes seems to be done..
Please check and let me know for further development. 

Thank you.

Controllers/HomeController.cs
```diff
@@ -3,21 +3,89 @@
...
+
+		static async Task<dynamic> CallApi(String path)
+		{
+			var response = await client.GetAsync(path);
+
+			if (!response.IsSuccessStatusCode)
+				return null;
+
+			var content = await response.Content.ReadAsStringAsync();
+			return JsonConvert.DeserializeObject(content);
```
MicahZoltu, line 79: Robustness: Rather than deserializing into a dynamic object, we should deserialize into a strongly typed class so that we get an error at deserialization time rather than access time.  This would also give auto-complete and compile-time checking of things.

Controllers/HomeController.cs
MicahZoltu: In this case, the performance hit isn't worth the complexity increase in my opinion, so I would prefer the null coalesce.  In some cases, performance wins are worth complexity increases while in others the performance wins aren't worth the complexity increases.  Async is a place where the performance gains are worth complexity increase because the potential gains from having a fully async call stack can be massive.  For this case, the additional cost of doing all of the matching doesn't outweigh the complexity increase of the extra conditional.

Controllers/HomeController.cs
```diff
@@ -3,21 +3,89 @@
...
+				var tagId = tagRegex.Match(path).Groups[1].Value;
+				var minPrice = minPriceRegex.Match(path).Groups[1].Value.TryParseUInt64();
+				var maxPrice = maxPriceRegex.Match(path).Groups[1].Value.TryParseUInt64();
+				var aboutUs = aboutRegex.Match(path).Success;
+
+				if (productId.HasValue)
+				{
+					var product = await CallApi($"products/{productId}");
+
+					url = $"https://bagcupid.com/{path}";
```
MicahZoltu, line 55: Bug: This should only happen if the product resolves.  If there is no product found with `productId` then we should leave the URL as `https://bagcupid.com/`.

MicahZoltu: Bug: If `path` contains a leading `/`, this URL will incorrectly be `https://bagcupid.com//app/...`  We should trim any leading `/` from the `path` before appending it here.  `$"https://bagcupid.com/{path.TrimLeadingSlash()}` (I think there is a trim function that lets you trim specific characters, if not you can create an extension function to do it like above).

Controllers/HomeController.cs
```diff
@@ -3,21 +3,89 @@
...
+
+					if (product?.images != null && product.images.Count > 0)
+						image = (string)(((JObject)product)["images"].Aggregate((minItem, nextItem) => (int)minItem["priority"] < (int)nextItem["priority"] ? minItem : nextItem))["large"];
```
MicahZoltu, line 58: Style: Can this be broken up into some helper functions to make the logic a little easier to follow?  Also, can we avoid some of the casting that is happening by having a strongly typed `product` object (see other comment)?  At the moment, this is a bit at risk of bugs due to the high complexity density of this code and the unchecked casts that are happening.

Also, if the chosen image doesn't have a `large` size this will either crash or end up with a `null` image (I'm not sure which).

Controllers/HomeController.cs
MicahZoltu: Bug: This needs to be `https://bagcupid.com/app/{path}`.  At the moment, the `/app/` segment is missing.

MicahZoltu (approved):
There are some comments in here but I will merge the PR and they can be addressed in a future PR.  I will be creating the infrastructure for unit tests after I merge this, so _MAKE SURE_ you sync up with my branch before continuing on with any work.

Controllers/HomeController.cs
inabeel: Got it :)

Controllers/HomeController.cs
inabeel: fixed.. added extension method

Controllers/HomeController.cs
inabeel: To break it, we probably need implementation of IComparable, which might be even more complex. So my recommendation is to be this function with comment to describe its purpose.

Casting is already eliminated in your defined type "Product", as conversion is held. When large will not be coming in JSON, it will have null value.

Controllers/HomeController.cs
inabeel: Seems you've already done this conversion to Custom Type

PR #66, opened 2016-11-21
Pull request for demanded changes and tests fixes

inabeel:


code/Controllers/HomeController.cs
```diff
@@ -42,24 +43,41 @@ public HomeController(BagsApi bagsApi)
...
 		}
 	}
 
+	public class MetaViewModel
+	{
+		public String url { get; set; }
+		public String type { get; set; }
+		public String title { get; set; }
+		public String description { get; set; }
+		public String image { get; set; }
```
MicahZoltu: Style: Public member variables/properties should have capital first letters in their name.

code/Controllers/HomeController.cs
```diff
@@ -25,11 +25,12 @@ public HomeController(BagsApi bagsApi)
 		[Route("app/{*path}")]
 		public async Task<IActionResult> App(String path = null)
 		{
-			ViewData["url"] = "https://bagcupid.com/";
-			ViewData["type"] = "website";
-			ViewData["title"] = "Bag Cupid";
-			ViewData["description"] = "What is your dream bag? Are you having trouble finding it? Let us help you!";
-			ViewData["image"] = "https://bagcupid.com/img/logo/bagcupid.png";
+			MetaViewModel model = new MetaViewModel();
```
MicahZoltu: Style: Use `var` over explicit type for variable declaration.

code/Controllers/HomeController.cs
```diff
@@ -42,24 +43,41 @@ public HomeController(BagsApi bagsApi)
...
 	{
 		public static UInt64? TryParseUInt64(this String value)
 		{
 			UInt64 id = 0;
 			return UInt64.TryParse(value, out id) ? id : (UInt64?)null;
 		}
+
+		public static String TrimLeadingSlash(this String value)
+		{
+			if(value.StartsWith("/"))
```
MicahZoltu: Style: This check is unnecessary, `TrimStart('/')` will trim all leading `/` off of the string, but if the string doesn't have a leading `/` character it will just return the original string.  Doing the `StartsWith` check adds complexity (along with this extension method) without providing value, so recommend just doing `path.TrimStart('/')`.

code/Controllers/HomeController.cs
```diff
@@ -42,24 +43,41 @@ public HomeController(BagsApi bagsApi)
...
 						.Aggregate((selectedImage, nextImage) => (nextImage.Priority < selectedImage.Priority) ? nextImage : selectedImage)
 						.Large;
 			}
 
-			return View("index");
+			return View("index", model);
 		}
 	}
 
+	public class MetaViewModel
```
MicahZoltu, line 61: Style: Nesting this inside of `HomeController` will give some context/clarity when instantiating it as to what it is a model for.  At the moment, we only have one controller, but if we add more controllers then having multiple objects named `ViewModel` will quickly lead to confusion as to which one is which.  By nesting inside of `HomeController` instantiating the model will look like `HomeController.ViewModel`, which is more clear as to what the view model is for.

That being said, the name should also include details about the action that the model is for.  In this case, it is for the `App` action so maybe we should name it `AppViewModel`?  Then instantiation would look like `HomeController.AppViewModel`, letting the reader immediately know what exactly this view model is used for.

Side note: What does the `Meta` mean in your name?

tests/Tests.cs
```diff
@@ -25,11 +25,13 @@ public class ControllerTests
 
 			// assert
 			var viewResult = Assert.IsType<ViewResult>(result);
-			Assert.Equal(expected: expectedDefaultUrl, actual: viewResult.ViewData["url"] as String);
-			Assert.Equal(expected: expectedDefaultType, actual: viewResult.ViewData["type"] as String);
-			Assert.Equal(expected: expectedDefaultTitle, actual: viewResult.ViewData["title"] as String);
-			Assert.Equal(expected: expectedDefaultDescription, actual: viewResult.ViewData["description"] as String);
-			Assert.Equal(expected: expectedDefaultImage, actual: viewResult.ViewData["image"] as String);
+			var viewModel = viewResult.Model as MetaViewModel;
```
MicahZoltu: Style: This should be an `Assert.IsType<MetaViewModel>(viewResult.Model)` so if the test fails, there is a very clear error indicating that the returned value isn't the right type.  By using `as MetaViewModel`, the error will be a Null Reference Exception instead of an invalid type exception, which is much less clear.

tests/Tests.cs
```diff
@@ -25,11 +25,13 @@ public class ControllerTests
...
+
+			Assert.Equal(expected: expectedDefaultUrl, actual: viewModel.url as String);
+			Assert.Equal(expected: expectedDefaultType, actual: viewModel.type as String);
+			Assert.Equal(expected: expectedDefaultTitle, actual: viewModel.title as String);
+			Assert.Equal(expected: expectedDefaultDescription, actual: viewModel.description as String);
+			Assert.Equal(expected: expectedDefaultImage, actual: viewModel.image as String);
```
MicahZoltu: Style: The `as String` can be removed from all of these now that we have a strongly typed model.

MicahZoltu (changes requested):
No major comments, just some minor style things.  Looks good overall.

code/Controllers/HomeController.cs
inabeel: I took Meta from Meta tags as this class is referring to Meta tags, and could be utilized wherever Meta Tags might be used of any sort.

code/Controllers/HomeController.cs
inabeel: fixed..

code/Controllers/HomeController.cs
inabeel: As it was intended to be general Meta Tag container, so do you want it to be changed?

code/Controllers/HomeController.cs
inabeel: fixed,,

code/Controllers/HomeController.cs
inabeel: fixed..

tests/Tests.cs
inabeel: fixed..

tests/Tests.cs
inabeel: fixed

code/Controllers/HomeController.cs
MicahZoltu: Ah, so the model is named after how it integrates with the view, not with how it integrates with the Controller/Action.  I like that convention now that I understand it.

Lets leave the class where it is, but change the name to `IndexViewModel`, since this is the view model for the entire Index page.  It happens to only contain details about the meta tags at the moment, but over time it may grow to include other parts of the index page and I don't think you can have multiple models per View?

code/Controllers/HomeController.cs
```diff
@@ -42,24 +43,38 @@ public HomeController(BagsApi bagsApi)
...
 		public static UInt64? TryParseUInt64(this String value)
 		{
 			UInt64 id = 0;
 			return UInt64.TryParse(value, out id) ? id : (UInt64?)null;
 		}
+
+		public static String TrimLeadingSlash(this String value)
+		{
+			return value.TrimStart('/');
+		}
```
MicahZoltu, line 81: Style: Recommend deleting this extension function.  I don't think `foo.TrimLeadingSlash` adds any clarity over `foo.TrimStart('/')`, plus the doc comments are all on `.TrimStart` and `TrimStart` is something that future developers are more likely to be familiar with in terms of behavior.  For example, this actually trims *all* leading slashes, e.g., `///app/product` but that isn't clear without the documentation on `TrimStart`.

code/Controllers/HomeController.cs
```diff
@@ -25,11 +25,12 @@ public HomeController(BagsApi bagsApi)
...
-			ViewData["type"] = "website";
-			ViewData["title"] = "Bag Cupid";
-			ViewData["description"] = "What is your dream bag? Are you having trouble finding it? Let us help you!";
-			ViewData["image"] = "https://bagcupid.com/img/logo/bagcupid.png";
+			var model = new MetaViewModel();
+			model.Url = "https://bagcupid.com/";
+			model.Type = "website";
+			model.Title = "Bag Cupid";
+			model.Description = "What is your dream bag? Are you having trouble finding it? Let us help you!";
+			model.Image = "https://bagcupid.com/img/logo/bagcupid.png";
```
MicahZoltu: Style: We should move all of these defaults into the view model initialization since every instance of the view model should always have them set to something, no matter where it is instantiated.
```
public class IndexViewModel
{
	public String Url { get; set; } = "https://bagcupid.com/";
	public String Type { get; set; } = "website";
	public String Title { get; set; } = "Bag Cupid";
	public String Description { get; set; } = "What is your dream bag? Are you having trouble finding it? Let us help you!";
	public String Image { get; set; } = "https://bagcupid.com/img/logo/bagcupid.png";
}
```

This will make it so future development can't accidentally forget to set one of these, and also if we end up having multiple controllers/actions that return the index view, we don't have to worry about ensuring they all share the same default values.

code/Controllers/HomeController.cs
inabeel: In case of Multiple View models, We can create Wrapper View Model class which could have MetaViewModel as it's child. This way Meta Tags will always be separated from other view logic

code/Controllers/HomeController.cs
inabeel: Fixed..

code/Controllers/HomeController.cs
inabeel: Fixed

MicahZoltu:
Closing in favor of #68

PR #73, opened 2016-12-10
Fix for Pinterest share

arpitux:
Required to replace https by http for the link in Pinterest share


code/client/scripts/index.js
```diff
@@ -1224,7 +1224,7 @@ function ShareLink(channel, entity, product_imgurl, product_name, product_brand)
                     pin_desc = pin_desc.trim().slice(0, -1)
                 }
             }
-            window.open('https://pinterest.com/pin/create/link/?url=' + encodeURIComponent(url) + '&media=' + encodeURIComponent(product_imgurl) + '&description=' + escape(pin_desc), "_blank");
+            window.open('https://pinterest.com/pin/create/link/?url=' + encodeURIComponent(url.replace("https","http")) + '&media=' + encodeURIComponent(product_imgurl) + '&description=' + escape(pin_desc), "_blank");
```
MicahZoltu: I don't believe this is the problem.  If you click the Pin button, it takes you to the following link:
```
https://www.pinterest.com/pin/create/link/?url=https%253A%2F%2Fbagcupid.com%2Fapp%2Ftags%2F3&media=https%3A%2F%2Fimages-na.ssl-images-amazon.com%2Fimages%2FI%2F51-LfhhdMNL._SL160_.jpg&description=%23style%3A%20clutch
```
If you URL decode everything after the `?url=` bit you get:
```
https%3A//bagcupid.com/app/tags/3&media=https://images-na.ssl-images-amazon.com/images/I/51-LfhhdMNL._SL160_.jpg&description=#style: clutch
```
The problem is that the `:` is double-encoded for some reason, but the rest of the URL isn't double decoded.

functional-tests/first_test.js.js
```diff
@@ -0,0 +1,10 @@
+var webdriver = require('selenium-webdriver');
+
+var driver = new webdriver.Builder().
+   withCapabilities(webdriver.Capabilities.chrome()).
+   build();
+
+driver.get('https://www.bagcupid.com');
+driver.findElement(webdriver.By.name('q')).sendKeys('simple programmer');
+driver.findElement(webdriver.By.name('btnG')).click();
+//driver.quit();
```
MicahZoltu: Can you remove this file from the PR until we are ready to merge it?

code/client/scripts/index.js
MicahZoltu: I did a bit of debugging and it appears that the `url` variable at this point in the code is `https%3A//bagcupid.com/app/tags/3`.  I'm not sure where the `url` variable is sourced from, but it is already partially encoded and it shouldn't be.  You will need to figure out what is doing the partial encoding and stop it to fix this bug.

MicahZoltu: Looks like the problem is up on line 1180, where you call `escape` on `window.location.href`.  I suspect the fix is to just do `url = window.location.href` on that line instead, since I don't believe the `.replace` is necessary anymore, and you shouldn't be escaping the URL at that place.

However, it appears there are other bugs that are going to crop up, in particular the other `case` blocks do not call `encodeURIComponent` before they encode the `url` but they should (just like the pinterest code does).

arpitux: I made the changes

functional-tests/first_test.js.js
arpitux: I removed it.

code/client/scripts/index.js
```diff
@@ -1207,24 +1207,23 @@ function ShareLink(channel, entity, product_imgurl, product_name, product_brand)
 
                 if (!(productId || tags || minPrice || maxPrice)){
                     //Sharing base webpage
-                    product_imgurl = window.location.origin + $("img.header-logo").attr("src");
-                    pin_desc = "Find Your Perfect Bag, We Make It Easy.";
+                    product_imgurl = encodeURI(window.location.origin + $("img.header-logo").attr("src"));
+                    pin_desc = escape("Find Your Perfect Bag, We Make It Easy.");
```
MicahZoltu: This should use `encodeURIComponent()` instead of `escape`.  A good description of which to use and when can be found here: http://stackoverflow.com/a/3608791/879046

code/client/scripts/index.js
```diff
@@ -1207,24 +1207,23 @@ function ShareLink(channel, entity, product_imgurl, product_name, product_brand)
...
-                   
+
                     var selected_tags = $("#main-search option:selected");
 
                     for (var i = 0; i < selected_tags.length; i++) {
-                       
                         pin_desc += selected_tags[i].text + ", ";
                     }
-                    pin_desc = pin_desc.trim().slice(0, -1)
+                    pin_desc = escape(pin_desc.trim().slice(0, -1));
```
MicahZoltu: Same as above, this should use `encodeURIComponent()`.

code/client/scripts/index.js
```diff
@@ -1175,9 +1175,9 @@ function ShareLink(channel, entity, product_imgurl, product_name, product_brand)
     var url = "", tag_names = "";
 
     if (entity == "product") 
-        url = escape(window.location.href.replace(window.location.pathname, "") + "/app/product/" + g_open_productid);
+        url = encodeURI(window.location.href.replace(window.location.pathname, "") + "/app/product/" + g_open_productid);
```
MicahZoltu: I believe all of these should be `encodeURIComponent()` as they are going to be used in a query string.  `encodeURI` won't encode the `://` correctly and will break the final URL that is constructed.

arpitux:
Changed encodeURI / escape with encodeURIComponent

MicahZoltu:
We should look into making sure we are escaping properly for the other social sharing buttons as well.

PR #74, opened 2016-12-15
Fixed single quote issue with social sharing

arpitux:


code/client/scripts/index.js
```diff
@@ -1544,4 +1545,8 @@ function ShowingBrandFilter() {
 
 function HideBanner() {
     $("html, body").stop().animate({ scrollTop: $(".banner").height() - 25}, '500');    
+}
+
+function EscapeSingleQuote(str) {
+    return str.replace(/'/g, "\\'");
```
MicahZoltu, line 1551: Single quote is one example out of a bunch of ways this could break.  Instead, we should be properly URL encoding or escaping any string that is passed around.  For example, what happens if there is a double quote in the name?  What about a `/` or a `\`?  Backtick?  etc.  Correctly escaping strings is quite complex which is why in generally you should *always* use one of the built-in library methods for encoding.

That being said, I'm not familiar enough with this code to know where the string in question is coming from or going to, so I'm not sure what the most appropriate way to escape it would be.

arpitux: Right, let me come up with better solution.

PR #75, opened 2016-12-16
Removed default design-v2 class

arpitux:


code/client/index.html
```diff
@@ -126,7 +126,7 @@
 	<link href="/vendors/owl-carousel/owl.carousel.css" rel="stylesheet" type="text/css" />
 	<link href="/vendors/owl-carousel/owl.theme.css" rel="stylesheet" type="text/css" />
 </head>
-<body class="design-v2" onload="ApplyDesignExperiemnt();">
+<body onload="ApplyDesignExperiemnt();">
```
MicahZoltu, line 129: Will this still allow the experiment to run, it will just default to no-experiment on-load and then snap to v2 once the experiment choice is fetched from Google?

arpitux: What experiment (JS code) do is, it first get the choice (either 0 or 1). For 0 it doesn't do anything. For 1 it adds "design-v2" class in body.

PR #76, opened 2016-12-18
Fixed product popup issue when brand is unavailable

arpitux:


code/client/index.html
```diff
@@ -126,7 +126,7 @@
 	<link href="/vendors/owl-carousel/owl.carousel.css" rel="stylesheet" type="text/css" />
 	<link href="/vendors/owl-carousel/owl.theme.css" rel="stylesheet" type="text/css" />
 </head>
-<body onload="ApplyDesignExperiemnt();">
+<body class="design-v2" onload="ApplyDesignExperiemnt();">
```
MicahZoltu: Didn't you just delete this in a previous PR?

arpitux: My bad. I actually keep this class added so that I see new design always while working locally.

arpitux: I removed it

PR #80, opened 2016-12-26
Products with tags, without product Id

inabeel:


code/BagsApi.cs
```diff
@@ -29,6 +30,39 @@ public BagsApi()
...
+			var url = new StringBuilder();
+			url.Append("products/by_tags?");
+
+			for (int i = 0; i < tags.Length; i++)
+			{
+				url.Append($"tag_id={tags[i]}");
+
+				if (i != tags.Length - 1)
+					url.Append("&");
+			}
```
MicahZoltu: This `for` loop could be greatly simplified by using `String.Join()`:
```c#
var url = $"products/by_tags?{String.Join("&", tags)}"
```

code/BagsApi.cs
```diff
@@ -29,6 +30,39 @@ public BagsApi()
 			return JsonConvert.DeserializeObject<Product>(content);
 		}
 
+		public virtual async Task<List<Product>> GetProductsByTags(String[] tags)
```
MicahZoltu: Versatility: Change `String[]` to `IEnumerable<String>`.  It is more versatile than `String[]` as you can pass in an array, a list, a set, etc.

code/Controllers/HomeController.cs
```diff
@@ -45,17 +46,42 @@ public HomeController(BagsApi bagsApi)
...
 			}
+			else if (!String.IsNullOrEmpty(tagId))
+			{
+				model.Url = $"https://bagcupid.com/app/{path.TrimStart('/')}";
+
+				var tags = tagId.Split('_');
+				var products = await bagsApi.GetProductsByTags(tags);
+
+				if (products?.Count > 0)
+					model.Images = products.Take(5)
```
MicahZoltu: Style: Put the `Take(5)` on its own line since you are chaining multiple methods here.

code/Controllers/HomeController.cs
```diff
@@ -45,17 +46,42 @@ public HomeController(BagsApi bagsApi)
...
+						.Select(product => product.Images
+							.OrderBy(image => image.Priority)
+							.SelectAndSwallowReferenceType(image => new Uri(image.Large))
+							.Where(image => image != null).FirstOrDefault());
```
MicahZoltu: Style: Put `FirstOrDefault()` on its own line since you are chaining multiple methods.

code/Controllers/HomeController.cs
```diff
@@ -45,17 +46,42 @@ public HomeController(BagsApi bagsApi)
...
+
+
+				var tagModels = await bagsApi.GetTags();
```
MicahZoltu: Performance: Move the call to `bagsApi.GetTags()` up to the top of the method and save the result in a Task (don't use `await`).  Then down here actually do the await.  This will make it so the query for this starts and is running in the background while we are also fetching the `ProductsByTags`.

Style: Also, we can get rid of one of the if statements along the way:

```c#
else if (...)
{
	model.Url = $"https://bagcupid.com/app/{path.TrimStart('/')}";
	var tags = tagId.Split('_');

	var productsTask = bagsApi.GetProductsByTags(tags);
	var tagModelsTask = bagsApi.GetTags();
	
	model.Images = (await productsTask ?? Enumerable.Empty<BagsApi.Product>())
		.Take(5)
		.Select(product => product.Images
			.OrderBy(image => image.Priority)
			.SelectAndSwallowReferenceType(image => new Uri(image.Large))
			.Where(image => image != null)
			.FirstOrDefault())
		.DefaultIfEmpty(model.Images.First());

	var tagModels = await tagModelsTask;
	...
}
```

The idea behind this is to wait until the last possible moment to `await` on something.  The more time you spend doing other work while waiting, the more efficient the code will be.  This is especially true if you have other network operations to do like in this case, where we need to fetch a couple different things.

code/Controllers/HomeController.cs
```diff
@@ -45,17 +46,42 @@ public HomeController(BagsApi bagsApi)
...
+
+				if (tagModels?.Count() > 0)
+				{
+					var tagsDescription = String.Join(", ", tagModels.Where(x => tags.Any(y => Convert.ToUInt32(y) == x.TagId))
+						.Select(x => x.TagName));
```
MicahZoltu: Style: Multiply nested LINQ statements can become really hard to read.  This one is particularly tricky so I recommend breaking it up into several separate steps with well named variables in-between.  Alternatively, it may be easier to just put all of the tags into a dictionary and then do a lookup by ID.

Readability: Use more verbose variable names (not `x` and `y`).

```c#
var tags = tagModels.ToDictionary(tagModel => tagModel.TagId, tagModel => tagModel.TagName);
var tagNames = tags.Select(tagId => tags[Convert.ToUInt32(tagId)]);
var tagsDescription = String.Join(", ", tagNames);
```

The first line extracts all of the tags into a dictionary for easy lookup.  The second line gets an enumerable of tag names from the IDs coming in from the query string.  The third line joins them all together.

tests/Tests.cs
```diff
@@ -83,6 +83,66 @@ public class ControllerTests
 		}
 
 		[Fact]
+		public async void valid_products_by_single_tag()
```
MicahZoltu: Add `_from_real_api` to the end of the name of this and the other test so it is clear that these tests don't use fake data.

tests/Tests.cs
```diff
@@ -83,6 +83,66 @@ public class ControllerTests
...
+			// assert
+			var viewResult = Assert.IsType<ViewResult>(result);
+			var viewModel = Assert.IsType<MetaViewModel>(viewResult.Model);
+
+			Assert.Equal(expected: "https://bagcupid.com/app/tags/190_303", actual: viewModel.Url);
+			Assert.Equal(expected: expectedDefaultType, actual: viewModel.Type);
+			Assert.Equal(expected: expectedDefaultTitle, actual: viewModel.Title);
+			Assert.Equal(expected: "Find your perfect adidas, flat handbag!", actual: viewModel.Description);
+			Assert.Equal(expected: expectedImages, actual: viewModel.Images);
+		}
```
MicahZoltu, line 143: Along with these two tests, I recommend adding some tests that use mock data (like the ones below) to test various failure scenarios.  Examples include:
* What happens if the API returns no tags?
* What happens if the API returns images that aren't valid URIs?
* What happens if the API returns some products with no images?
* What happens if the API returns all products with no images?
* What happens if the API returns tags with no name?

Generally speaking, the code should be robust against the API misbehaving and the code in this class should never completely fail in the face of bad data.  It should always fall-back to reasonable default results, and the tests should show that.

code/BagsApi.cs
inabeel: Done..

code/BagsApi.cs
inabeel: Done..

code/Controllers/HomeController.cs
inabeel: Done..

code/Controllers/HomeController.cs
inabeel: Done..

code/Controllers/HomeController.cs
inabeel: Done..

code/Controllers/HomeController.cs
inabeel: Done...

tests/Tests.cs
inabeel: Done..

code/Controllers/HomeController.cs
inabeel: You haven't added filter in Lambda to get only selected names.
It is eventually bringing all tag names

inabeel: it is fixed

code/Controllers/HomeController.cs
inabeel: var productsTask = bagsApi.GetProductsByTags(tagList);
                var taskTagModels = bagsApi.GetTags();

                var products = await productsTask;
                if (products?.Count > 0)
                    model.Images = (await productsTask ?? Enumerable.Empty<BagsApi.Product>())
                        .Take(5)
                        .Select(product => product.Images
                                .OrderBy(image => image.Priority)
                                .SelectAndSwallowReferenceType(image => new Uri(image.Large))
                                .Where(image => image != null)
                                .FirstOrDefault());

Avoiding this if statement will bring empty enumeration if there are no products available. This will update Model.Images to be replaced as well. which should not be the case

MicahZoltu: The `DefaultIfEmpty(model.Images.First())` takes care of dealing with the empty enumeration case.  It could be improved a little bit with a new extension method that accepted an enumerable so you could do `DefaultIfEmpty(model.Images)`.

code/Controllers/HomeController.cs
MicahZoltu: Oops, my mistake, I re-used a variable name in my example:
```c#
var tagIdsToNames = tagModels.ToDictionary(tagModel => tagModel.TagId, tagModel => tagModel.TagName);
var tagNames = tagIdsFromQueryString.Select(tagId => tagIdsToNames[Convert.ToUInt32(tagId)]);
var tagsDescription = String.Join(", ", tagNames);
```

code/Controllers/HomeController.cs
inabeel: The problem is with assignment, whether you assign something or left it on default value. As per my understanding towards your statement is that DefaultIfEmpty will assign default value if enumeration is empty. And Default could be image collection with no item or null. While we need one which was assigned in constructor.

So that is why i don't understand here the meaning of another extension method, which presumes to be additional complex step,

code/Controllers/HomeController.cs
inabeel: Already done 👍

tests/Tests.cs
inabeel: I have added more tests, and i feel these are serving these purposes now.
Please have a look at them, and let me know for any changes.

MicahZoltu:
It appears that several of the files in this PR have reverted back to spaces for indentation instead of tabs.  Can you fix that so it is easier to review the actual changes?

code/Controllers/HomeController.cs
```diff
@@ -5,79 +5,110 @@
...
+            {
+                model.Url = $"https://bagcupid.com/app/{path.TrimStart('/')}";
+
+                var tagList = tagIds.Split('_').ToList();
+                var productsTask = bagsApi.GetProductsByTags(tagList);
+                var taskTagModels = bagsApi.GetTags();
+
+                var products = await productsTask;
+                if (products?.Count > 0)
+                    model.Images = (await productsTask ?? Enumerable.Empty<BagsApi.Product>())
```
MicahZoltu: This should be `model.Images = (products ?? Enumerable.Empty<BagsApi.Product>())` since you already have awaited the task above.

Also, you can avoid dealing with nulls by moving the null coalesce up a bit:
```
var products = (await productsTask) ?? Enumerable.Empty<BagsApi.Product>();
if (products.Count > 0)
   ...
```

Lastly, because the code inside the if block is lengthy, it should include curly braces to make it clear where the beginning/end is.

inabeel: i forgot to replace later line with products. We cant use that statement as it will assign null to model.Images and we will lose our default value.

PR #85, opened 2017-01-18
Added UI Tests (Nightwatch)

arpitux:


functional-tests/nightwatch.json
```diff
@@ -4,11 +4,11 @@
...
 			"selenium_host": "selenium-hub",
-			"screenshots": {
-				"enabled": true,
-				"on_failure": true,
-				"path": "tests_output"
-			}
+      "screenshots": {
+        "enabled": true,
+        "on_failure": true,
+        "path": "C:/FL/MicahZoltu/HandbagsProject/MicahRepo/bags-client/functional-tests/reports"
```
MicahZoltu: This should be a relative path.  I don't, for example, have `C:/FL/MicahZoltu` on my computer.  Also, in linux (where this is run) that isn't a valid path.

MicahZoltu:
The files appear to have mixed tabs and spaces for indentation.  Can you make it so all lines have the same indentation so it renders correctly across all editors and diffs are easy to read?

Personally, I prefer tabs for indent in all files that support it (which includes `.js` and `.json`) though I'm not sure what the indentation scheme is for other `.js` files in this project are.  I believe `.json` files in the project are generally indented with tabs.

Exactly how to fix this depends on your editor.  If you are using Visual Studio, you can view white space with `Edit (menu) > Advanced (submenu) > View White Space (menu item)`.  This will show you which lines have tabs and which have spaces.  You can then select all in the file and do `Edit (Menu) > Advanced (submenu) > Tabify Selected (menu item)`.

Be aware though, it appears that there isn't currently a consistent indentation strategy, so you may need to do some additional cleanup afterwards.

functional-tests/nightwatch.json
```diff
@@ -4,11 +4,11 @@
...
+      }
```
MicahZoltu: The rest of this file is `tab` indented, can you change these lines to match?

functional-tests/tests/tests.js
```diff
@@ -1,8 +1,118 @@
...
-	'Nightwatch.js Test': function (browser) {
-		browser
-		  .url(process.env.BAG_CUPID_URL)
-		  .assert.title('Bag Cupid')
-		  .end();
-	}
+	'Check for page title and Banner': function (browser) {
+		browser.url(process.env.BAG_CUPID_URL);
+		browser.assert.title('Bag Cupid');
+		browser.pause(6000);
```
MicahZoltu, line 5: We can get this merged as-is with these `pause`s everywhere, but as a next step we should try to remove them by using nightwatch's wait until syntax.  At the moment, some of these tests take almost 20 seconds when they really should run in under a second (assuming fast network).  I have some changes I want to do in the near future to make running the tests locally much faster as well, but they won't help if we are using `pause` instead of the `wait until` stuff.

Lets leave that to a follow-up Pull Request though, for now just address the other couple minor comments.

functional-tests/nightwatch.json
arpitux: Okay

functional-tests/tests/tests.js
arpitux: I kept longer pauses considering my slower network :)

arpitux: If I keep short pauses some of the test are then more likely to get failed

functional-tests/nightwatch.json
arpitux: Done

PR #86, opened 2017-01-30
Changes in UI Tests

arpitux:
Used waitForElementNotPresent, waitForElementNotVisible, waitForElementPresent, waitForElementVisible instead of pause() commands


functional-tests/tests/tests.js
```diff
@@ -25,68 +23,60 @@ module.exports = {
 		browser.end();
 	},
 	'Check Tag in search box for URL with tag filter': function (browser) {
-		browser.url("http://bagcupid.com/app/tags/3")
-		browser.pause(7000);
-		browser.assert.visible(".select2-container--default .select2-selection--multiple .select2-selection__choice");
+	    browser.url("http://bagcupid.com/app/tags/3");
+	    browser.waitForElementVisible(".select2-container--default .select2-selection--multiple .select2-selection__choice", 7000);
```
MicahZoltu: These two lines have mixed tabs/spaces.  Most editors have a way of alerting you when this happens, which do you use?

functional-tests/tests/tests.js
```diff
@@ -25,68 +23,60 @@ module.exports = {
...
 		browser.end();
 	},
-	'Banner should be hidden for URL with filter': function (browser) {
-		browser.url("http://bagcupid.com/app/tags/3");
-		browser.pause(7000);
+	'Side filter should open': function (browser) {
+	    browser.url(process.env.BAG_CUPID_URL);
+		browser.waitForElementVisible("#side-filter-trigger",7000);
 		browser.click("#side-filter-trigger");
 		browser.pause(2000);
```
MicahZoltu: Looks like you missed a pause here.

MicahZoltu:
I just notice this https://github.com/Zoltu/bags-client/pull/86/files#diff-cd90f9f9bb15a04809bb805a131b23cdR95.  The prices are automatically updated to match Amazon, which means they change with time.  I recommend against having tests that validate the price is a certain value and instead just make sure that there *is* a price and it is a positive number or something.

functional-tests/tests/tests.js
arpitux: I use visual studio. I pushed the change..

functional-tests/tests/tests.js
arpitux: Here I added a test on "side bar filter" to ensure it opens when clicking on Filter button in header. I have to use `pause` here because I can't use any other alternative here. Nothing else happens on UI apart from the side bar's `style` get changed

arpitux:
Sorry, I didn't understand your last comment.

> I just notice this https://github.com/Zoltu/bags-client/pull/86/files#diff-cd90f9f9bb15a04809bb805a131b23cdR95. The prices are automatically updated to match Amazon, which means they change with time. I recommend against having tests that validate the price is a certain value and instead just make sure that there is a price and it is a positive number or something.

MicahZoltu:
On line 95 you assert that the price of a bag is $800.  This will likely break at some point because the prices change over time.  There is a background job that is always running syncing the prices of bagcupid.com with the prices listed on Amazon.com and it is very likely that the bag you are testing will have a change in price at some point and the tests will start failing.  The test should not assert a specific price is set, only that there is a price set and it is valid (all numbers, etc.)

MicahZoltu:
I'm getting an error on one test:
```
 chrome   ✖ Timed out while waiting for element <.enjoyhint-step-1> to be present for 4000 milliseconds.  - expected "visible" but got: "not found"
 chrome       at Object.Check for Help tour (/app/tests/tests.js:62:11)
              at _combinedTickCallback (internal/process/next_tick.js:67:7)
 chrome   FAILED:  1 assertions failed and 1 passed (6.987s)
 chrome
 chrome    _________________________________________________

 TEST FAILURE:  1 assertions failed, 15 passed. (47.346s)
 chrome    ✖ tests

   - Check for Help tour (6.987s)
   Timed out while waiting for element <.enjoyhint-step-1> to be present for 4000 milliseconds.  - expected "visible" but got: "not found"
       at Object.Check for Help tour (/app/tests/tests.js:62:11)
       at _combinedTickCallback (internal/process/next_tick.js:67:7)
```

It doesn't appear to fail every time, though it does fail for at least one of the two browsers each run.  My network connection should be very good from here, so the 4 second timeout should be plenty.  It seems that maybe it fails the first time, does this test rely on a particular cookie being set or unset?  We may need to tell the test to clear cookies before executing, to make sure the test runs as a first-time user if that is necessary.  I suspect nightwatch has a way to do this?

arpitux:
For line number 95 I am testing for this url.
https://bagcupid.com/app/minprice/100/maxprice/800,
to insure that the price range which we display on header is correct. This test is not for a specific Bag.

arpitux:
I am looking into the test which is failing..

arpitux:
I just executed these tests many times.
I found that one random test fails for firefox only. For chrome all the tests get passed.
So it's not like only once specific test is failing all the time. The error for failed test is always a Timeout error.

PR #90, opened 2017-05-05
Pinterest sharing issue

arpitux:
Fixed the issue for product name with space in pinterest sharing

### DIFF
--- a/code/client/scripts/index.js
+++ b/code/client/scripts/index.js
@@ -1182,7 +1182,7 @@ function ShareLink(channel, entity, product_imgurl, product_name, product_brand)
         case 'pinterest':
             var pin_desc = "";
             if (entity == "product")
-                pin_desc = encodeURIComponent((product_brand != "" ? product_brand + ": " : "")) + product_name;
+                pin_desc = (product_brand != "" ? product_brand + ": " : "") + product_name;
             else if (entity == "search") {
                 var path = window.location.pathname;
 


code/client/scripts/index.js
```diff
@@ -1182,7 +1182,7 @@ function ShareLink(channel, entity, product_imgurl, product_name, product_brand)
         case 'pinterest':
             var pin_desc = "";
             if (entity == "product")
-                pin_desc = encodeURIComponent((product_brand != "" ? product_brand + ": " : "")) + product_name;
+                pin_desc = (product_brand != "" ? product_brand + ": " : "") + product_name;
```
MicahZoltu, line 1185: I don't think this resolves the issue as `product_brand` is already URL encoded before it is passed to this function (`ShareLink`).

If you inspect the pinterest button on a product you will see:
```html
<button class="btn bgm-pinterest btn-lg dropdown-toggle social-share-button pinterest waves-effect" type="button" onclick="ShareLink('pinterest', 'product','https://images-na.ssl-images-amazon.com/images/I/51M4Q5b-nKL.jpg','dillon','michael%20kors');">
```
Notice `ShareLink(..., 'michael%20kors');`

This tells me that the problem is somewhere else and that the brand is being encoded before it gets to this line of code.

arpitux: Yes it's already encoded so that's why I removed further encoding. Didn't this solution work?

